PR #33, opened 2020-08-09
Fix ellipse tool

Equbuxu:
Ellipse tool now works like it is supposed to


flabbet:
Works like a charm, awesome job!

PR #36, opened 2020-08-15
Zoom tool

flabbet:
Added Zoom tool and fixed editable textbox shortcut bug.


flabbet:
@Equbuxu Could you check if it works well with a stylus?

Equbuxu:
It does work with stylus about as well as it does with mouse. It works, but it will definitely require improvements in the future since:
1. It feels quite laggy, I believe this because the Use() function of the tool accepts integer coordinates instead of float or double
2. It should zoom towards mouse down position on the canvas, not towards center
3. In my opinion, it could be more sensetive

flabbet:
It doesn't use integer `Use()` coordinates, I calculate drag with absolute screen position. I'll try to improve it.

flabbet:
@Equbuxu Can you check now?

flabbet:
Let me describe how it works: https://www.youtube.com/watch?v=YPN0qhSyWy8

flabbet:
@Equbuxu Can you check now :)?

flabbet:
Okay, merging it. #39 yet to be fixed in the future

PR #42, opened 2020-09-07
Add a tool size increase and decrease shortcuts using [ and ]

brendan-holmes:
- Added new shortcut bindings for '[' and ']' in View Model with -1 and +1 parameter for decrease and increase respectively.
- Created new command 'ChangeToolsizeCommand' which executes a new method called 'ChangeToolsize'.
- Added setter for 'Toolsize' property in BitmapManager.
- 'ChangeToolsize(object parameter)' increments/decrements the current value of 'Toolsize' by the value of 'parameter' if the resultant value is > 0.

Tests:
image

### DIFF
--- a/PixiEditor/Models/Controllers/BitmapManager.cs
+++ b/PixiEditor/Models/Controllers/BitmapManager.cs
@@ -12,6 +12,7 @@ using PixiEditor.Models.ImageManipulation;
 using PixiEditor.Models.Layers;
 using PixiEditor.Models.Position;
 using PixiEditor.Models.Tools;
+using PixiEditor.Models.Tools.ToolSettings;
 
 namespace PixiEditor.Models.Controllers
 {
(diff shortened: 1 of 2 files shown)


PixiEditor/Models/Controllers/BitmapManager.cs
```diff
@@ -43,9 +43,16 @@ public Layer PreviewLayer
...
-        public int ToolSize => SelectedTool.Toolbar.GetSetting("ToolSize") != null
-            ? (int) SelectedTool.Toolbar.GetSetting("ToolSize").Value
+        public int ToolSize
+        {
+            get => SelectedTool.Toolbar.GetSetting("ToolSize") != null
+            ? (int)SelectedTool.Toolbar.GetSetting("ToolSize").Value
             : 1;
+            set
+            {
+                SelectedTool.Toolbar.GetSetting("ToolSize").Value = value;
```
flabbet, line 53: Using shortcuts on the tool, which doesn't have `ToolSize` raises an error, you should add null checking on the setter.

flabbet (changes requested):
Also, it would be awesome if the preview highlight changed in real-time, not only after moving the mouse

flabbet (approved):
Awesome, now it works great

brendan-holmes:
Thanks for the review! Glad to have done my first-ever PR on this fun project!

PR #46, opened 2020-09-12
Clear Selection when single point is selected. Add MaximiseWindow Shortcut. Add Exit to File Menu

PolarJunction:
## Pull Requests

### Modifications
* Updated SelectTool to clear the selection when a single point is clicked. The tool feels much more natural to clear by not selecting anything. It would be nice in the future to also clear the selection when we don't click on the canvas.
* Added Exit (CloseWindow) command to the File Menu, I think most applications have the ability to exit this way.
* Added MaximiseWindow (F11) shortcut.

image

#### Select Tool
* Create a new file
* Make a selection of a group of points
* Select a single point
* **VERIFY** that the selection is cleared.

#### Shortcuts
* Create a new file, and add some pixels
* Select File>Exit
* **VERIFY** that the CloseWindow confirmation is displayed
* Press cancel, then reduce the size of the Window
* Press F11
* **VERIFY** that the Window is maximized.

#### Test Cases

* All test cases passing, noticed intermittent failures with the ClipboardControl tests.

### DIFF
--- a/PixiEditor/Views/MainWindow.xaml
+++ b/PixiEditor/Views/MainWindow.xaml
@@ -82,6 +82,7 @@
                     <MenuItem Header="_Save As..." InputGestureText="Ctrl+Shift+S"
                               Command="{Binding SaveDocumentCommand}" CommandParameter="AsNew" />
                     <MenuItem Header="_Export" InputGestureText="Ctrl+Shift+Alt+S" Command="{Binding ExportFileCommand}" />
+                    <MenuItem Header="_Exit" InputGestureText="Ctrl+Q" Command="{x:Static SystemCommands.CloseWindowCommand}" />
                 </MenuItem>
                 <MenuItem Header="_Edit">
                     <MenuItem Header="_Undo" InputGestureText="Ctrl+Z" Command="{Binding UndoCommand}" />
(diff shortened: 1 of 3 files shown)


PixiEditor/Views/MainWindow.xaml
```diff
@@ -82,6 +82,7 @@
                     <MenuItem Header="_Save As..." InputGestureText="Ctrl+Shift+S"
                               Command="{Binding SaveDocumentCommand}" CommandParameter="AsNew" />
                     <MenuItem Header="_Export" InputGestureText="Ctrl+Shift+Alt+S" Command="{Binding ExportFileCommand}" />
+                    <MenuItem Header="_Exit" Command="{x:Static SystemCommands.CloseWindowCommand}" />
```
flabbet, line 85: One small thing I would add, a `<Separator />` above the Exit menu option.

flabbet (changes requested):
About the Maximize shortcut, I think it would make more sense if it would toggle the actual fullscreen, that covers the Windows taskbar too. To do that you will need to change the max-height probably, check `MainWindow.cs`.

flabbet:
Everything else good 💯

PolarJunction:
Added separator to exit menu.

Wasn't able to add Full Screen toggle. the following options are required to be set to achieve Full Screen:

-  MaxHeight = SystemParameters.PrimaryScreenHeight;
-  ResizeMode = ResizeMode.NoResize;
-  WindowState = WindowState.Maximized;
-  WindowStyle = WindowStyle.None;

Recommend a separate issue is raised to create a toggle command and replace the F11 functionality.

flabbet (approved):

Issue #44, opened 2020-09-11
Change object based settings to typed generics


flabbet:
**Is your feature request related to a problem? Please describe.**
Using tool Settings requires to manually cast `object` to the requested type, this is not comfortable to use and might make it more fragile to errors.

**Describe the solution you'd like**
Change `Setting` class to `Setting<T>` where `T` is a value type of setting. Currently to get `ToolSize` setting we need to do `(int) Toolbar.GetSetting("ToolSize").Value` but something like `Toolbar.GetSetting<int>("ToolSize").Value` would be better.

anilkhandei:
@flabbet  I am trying to convert settings class to a generic class. I wonder how to handle the LoadSharedSettings that is defined in the Toolbar class, any tips?

flabbet:
The simplest solution I can think of is that the top parent class would be the current implementation of `Setting`, that operates on objects, and then write a wrapper class that implements generic `T` type, like that
```csharp
class Setting<T> : Setting
```

After you do that, you will still have backward compatibility, so you probably won't need to change anything in the `Toolbar` class.

ArtemK123:
Hi, I would like to try this one. Could you please assign the issue to me?

ArtemK123:
Here is a pull request for the issue
https://github.com/PixiEditor/PixiEditor/pull/66

flabbet:
Closed in #66

PR #51, opened 2020-09-19
#44

anilkhandei:
## Pull Requests

Added: `Setting<T>` class
Modified: 
1. `Setting GetSetting` to `Setting<T> GetSetting<T>`
2. Updated all references of `GetSetting` method to use new 'GetSetting<T>' method

 Pull request rules:

### DIFF
--- a/PixiEditor/Models/Controllers/BitmapManager.cs
+++ b/PixiEditor/Models/Controllers/BitmapManager.cs
@@ -46,12 +46,12 @@ namespace PixiEditor.Models.Controllers
 
         public int ToolSize
         {
-            get => SelectedTool.Toolbar.GetSetting("ToolSize") != null
-            ? (int)SelectedTool.Toolbar.GetSetting("ToolSize").Value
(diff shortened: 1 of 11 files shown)


flabbet:
Please fix tests :)

anilkhandei:
I forgot to checkin the same. Please re validate.

pixitests_44

flabbet:
This is okay, but much better solution would be to provide a type of setting class, instead of value type. Because we know the setting we want to get, not necessarily the type it returns. 

For example instead `GetSetting<int>("ToolSize")` would be `GetSetting<SizeSetting>("ToolSize")

And return type would be `SizeSetting`

Could you change `GetSetting<T>` so `T` is a child class of `Setting`? And create a overload for `GetSetting` that doesn't need `T`?

ArtemK123:
Ok, I understand what you want.
I will do it in a couple of days, just wait for the updates ;-)

ArtemK123:
Hi. Maybe it is better to close that pull request, cause you have already merged my pull request related to the same issue.

flabbet:
Closed in #66

Issue #49, opened 2020-09-16
Clicking "Save" or "Save As..." and canceling, results in false "saved" state


flabbet:
**Describe as detailed as possible when it happens**
While having unsaved changes, click the "Save" or "Save As..." option (or using shortcuts) and closing the dialogs (without actually saving the file) program marks changes as saved, so closing the program does not trigger "Are you sure?" dialog.

**Potential fix***
The fastest solution would be to just check if the file was saved successfully, and set the value of `_unsavedDocumentModified` in `ViewModelMain` corresponding to the result.

PR #54, opened 2020-09-29
Resolved issue #49

hackf5:
Added a return value to `Exporter.SaveAsEditableFileWithDialog` indicating whether the dialog performed the save (true) or not (false).

Unfortunately it's not possible to add a meaningful test here because there is no abstraction around the file save dialog, and adding that abstraction and making it configurable in the test assembly seems to go beyond the scope of the bug.


flabbet:
Nice work

Issue #53, opened 2020-09-29
Pen up is not detected when pen up occurs outside the application


hackf5:
1. Using the pen tool draw a line and release the mouse button below the bottom of the screen onto the windows task bar.
2. Hold the mouse button down to the right of and off the canvas and drag the mouse onto the canvas.

A diagonal line appears between where you exited the canvas at the bottom and where you entered the canvas on the right. In the image below the green pixel is where I exited the canvas and released the mouse button and the red pixel is where I entered the canvas.

![image](https://user-images.githubusercontent.com/48881437/94579000-40b30980-0270-11eb-8139-4e982ec679fb.png)

flabbet:
Closed in #55

PR #55, opened 2020-09-29
Resolved issue #53

hackf5:
Added a global mouse hook to listen for all mouse up events so that if mouse up occurs outside of the application while drawing it is honoured. Standard WPF is oblivious to this event.

This might look a bit heavy duty, but it's the only way to detect a mouse up event outside of the application. It is actually a standard approach in many WPF applications. I've used this technique a number of times before and this implementation looks pretty solid.


flabbet:
This is a very solid solution, well done. Thanks for contributing!

hackf5:
Pleasure.

Issue #48, opened 2020-09-13
Draw nice pixel art to display in README


flabbet:
# Description
Draw a beautiful pixel art in PixiEditor so we can display it on the main page (README.md). Everyone visiting the repository (and website in the future) will see your work in the first place!

# Requirements
- A full screenshot with software open, not an exported image. Check current [README](https://github.com/PixiEditor/PixiEditor/blob/master/README.md) for reference.
- .pixi file
- Image **must** be drawn in PixiEditor

_By submitting you agree to give us a full right to all files you send._

flabbet:
#64 closed

PR #64, opened 2020-10-06
Adding new art for README (#48)

crhaugen:
I took a shot at creating an art piece for the README. 

If it's not what you are looking for, feel free to say so.

I wasn't sure how I should do the pull request so I replaced the old .png with my new one (same name). 

I am also attaching the .png to this comment. Github won't accept .pixi format in the comment so if there is some other way you would like me to add that one, please let me know. 

Screenshot


flabbet:
Hey, this looks cool! However would be much better without empty pixels around the image and Visual studio debug bar, you can just use Document -> Clip canvas. Please send us .pixi file on our [artwork channel](https://discord.gg/zACaVxU)

crhaugen:
Ok, I added the requested changes and made the image a bit bigger also. If there is anything else, please let me know. 

Screenshot

PR #66, opened 2020-10-08
Change object based settings to typed generics #44

ArtemK123:
## Pull Requests

 Pull request rules:

I created the Setting<T> class as an inheritor of Setting. I moved the Value property and associated field to the Setting<T> in order to have the generic Value.

I inherited other settings from Setting<T> where possible.

I added Toolbar.GetSettings<T> method in order to get a generic Setting where it is required. In addition, I used this method in places where the (some type) casts were placed.

One of the biggest problems was to refactor the SaveToolbarSettings and LoadSharedSettings methods. They were using the Setting.Value property but all the settings are cast to the base class. So, there is no information about the type of value and it breaks the idea of generic usage. I got rid of using Value property - instead, I just re-insert the record to the collection if this record was already inserted.

Also, I covered the Toolbar.GetSettings<T> by tests. Maybe, some more tests are required- it is ok to write to me about it, I will try to cover as much of my changes as needed.

During the manual testing, I have gotten the Null-pointer exception couple of times, though all tests were green. I have tried to find and fix all of those NPE places, but I still might miss some of them.

I think the usage of Toolbar.GetSetting<T>(settingName) is the best example, you can find some of them in changed files

I haven`t added any dependencies

All tests are green on my machine.

### DIFF
--- a/PixiEditor/Models/Tools/ToolSettings/Toolbars/Toolbar.cs
+++ b/PixiEditor/Models/Tools/ToolSettings/Toolbars/Toolbar.cs
@@ -22,15 +22,33 @@ namespace PixiEditor.Models.Tools.ToolSettings.Toolbars
         }
 
         /// <summary>
+        ///     Gets setting with given type T in toolbar by name.
+        /// </summary>
+        /// <param name="name">Setting name, non case sensitive</param>
+        /// <returns></returns>
+        public Setting<T> GetSetting<T>(string name)
+        {
+            Setting setting =  Settings.FirstOrDefault(currentSetting => string.Equals(currentSetting.Name, name, StringComparison.CurrentCultureIgnoreCase));
+
+            if (setting == null)
+                return null;
+
+            if (setting is Setting<T> convertedSetting)
+                return convertedSetting;
+
+            throw new Exception("Setting has value with unexpected type");
+        }
+
+        /// <summary>
         ///     Saves current toolbar state, so other toolbars with common settings can load them.
         /// </summary>
         public void SaveToolbarSettings()
         {
-            for (int i = 0; i < Settings.Count; i++)
-                if (SharedSettings.Any(x => x.Name == Settings[i].Name))
-                    SharedSettings.First(x => x.Name == Settings[i].Name).Value = Settings[i].Value;
-                else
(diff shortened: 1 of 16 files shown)


ArtemK123:
One of the checks was failed because of "'codecov' is not recognized as an internal or external command,
operable program or batch file." in "Upload to Codecov.io" step. Could you please help me to fix the problem?

flabbet:
Don't worry about codecov, it's just a pipeline bug.

ArtemK123:
So, should I add something else before the merging?

PixiEditor/Models/Tools/ToolSettings/Toolbars/Toolbar.cs
```diff
@@ -20,26 +21,46 @@ public virtual Setting GetSetting(string name)
...
+        /// <returns></returns>
+        public Setting<T> GetSetting<T>(string name)
+        {
+            Setting setting =  Settings.FirstOrDefault(currentSetting => string.Equals(currentSetting.Name, name, StringComparison.CurrentCultureIgnoreCase));
+
+            if (setting == null || !(setting is Setting<T> convertedSetting))
+                return null;
+
+            return convertedSetting;
+        }
```
flabbet, line 37: In my opinion would be better to insert in `T` the `Setting` based class, not the `Value` type
So instead of `Setting<T>` it would return `T` Where `T` is a class that inherits from `Setting`. I described this more detailed here https://github.com/PixiEditor/PixiEditor/pull/51#issuecomment-695974432

flabbet (changes requested):

ArtemK123:
I have done the changes, and all tests were green on my machine. But during the Azure check, the UndoManagerTests test case was failed. Should I somehow fix it or is it a known flaky test?

flabbet:
It's a flaky test, not sure it fails sometimes, don't worry about it.

ArtemK123:
Please, review the change-request. I have changed the GetSetting<T> in order to pass T as an inheritor of the Setting. I hope, I have caught the idea rightly.

Feel free to comment/ask questions

flabbet:
I have nothing to complain about, great job! 👍

flabbet (approved):

Issue #67, opened 2020-10-09
Crashes when opening invalid file


CPKreuz:
At the moment PixiEditor crashes when trying to open a invalid/corrupted file

It would be good if it would display a error message instead

PR #70, opened 2020-10-09
Handle corrupt file open exception

pauliax:
## Pull Requests

 Pull request rules:

solves #67 

show an error instead of silent crash:
image

sure

checked - all ok


flabbet (changes requested):
Hey, thanks for contributing! I tested a bit your solution and it doesn't really work. You wrapped the ShowDialog in the try-catch clause, but it is not a place where deserialization happens. In some places, code directly calls `Importer.ImportDocument`. This is where you should look.

PR #73, opened 2020-10-10
Optimize FloodFill

### LINKED ISSUE #34: Optimize flood fill
Here is a comment part about Flood Fill from [Reddit post](https://www.reddit.com/r/csharp/comments/hjx9gn/hey_after_2_years_of_development_i_published_the/fwplg7l?utm_source=share&utm_medium=web2x)

# FloodFill
You make a deep copy of the layer you flood fill on. That's an expensive operation right of the bat.

You use GetPixel and SetPixel on the clone to keep track of what pixels you have changed. That's a really expensive way of doing it. Consider replacing it with a bool[,] array with the same size as the layer. Then you can remove the clone and just sample from the original layer.

Consider checking if you've already seen a pixel before you put it into the stack. You could use the bool[,] array to keep track of pixels you've already placed into the stack and then check it before you place neighboring pixel Coordinates into the stack. It keeps the stack smaller and you have to do those checks later anyway.

Change the order of the neighbors you put into the stack. Adding x + 1 and x - 1 last, makes your data access a little more linear which increases the chances of the data being in the cache.

brogowski:
Resolves #34 

I've run test benchmark on 1024x1024 bitmap with a circle of 384px radius as fill target.

All the changes mentioned in Reddit post are implemented.

The only other significant performance boost would be to parallelize bitmap traversal, but since in WPF bitmaps have thread affinity, it would require cloning whole bitmap...

### Before

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1082 (1909/November2018Update/19H2)
AMD Ryzen 7 3700X, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=3.1.402
  [Host]        : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
  .NET Core 3.1 : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT

Job=.NET Core 3.1  Runtime=.NET Core 3.1  

```
|        Method |    Mean |    Error |   StdDev |     Min |     Max |  Median |
|-------------- |--------:|---------:|---------:|--------:|--------:|--------:|
| FloodFillTest | 1.298 s | 0.0038 s | 0.0036 s | 1.291 s | 1.304 s | 1.298 s |

### After

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1082 (1909/November2018Update/19H2)
AMD Ryzen 7 3700X, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=3.1.402
  [Host]        : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
  .NET Core 3.1 : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT

Job=.NET Core 3.1  Runtime=.NET Core 3.1  

```
|        Method |    Mean |    Error |   StdDev |     Min |     Max |  Median |
|-------------- |--------:|---------:|---------:|--------:|--------:|--------:|
| FloodFillTest | 1.016 s | 0.0064 s | 0.0057 s | 1.005 s | 1.027 s | 1.016 s |

### DIFF
--- a/PixiEditor/Models/Tools/Tools/FloodFill.cs
+++ b/PixiEditor/Models/Tools/Tools/FloodFill.cs
@@ -24,8 +24,8 @@ namespace PixiEditor.Models.Tools.Tools
         {
             List<Coordinates> changedCoords = new List<Coordinates>();
 
-            int width = layer.Width;
-            int height = layer.Height;
+            int width = ViewModelMain.Current.BitmapManager.ActiveDocument.Width;
+            int height = ViewModelMain.Current.BitmapManager.ActiveDocument.Height;
 
             var visited = new bool[width, height];
 
--- a/PixiEditor/Models/Tools/Tools/FloodFill.cs
+++ b/PixiEditor/Models/Tools/Tools/FloodFill.cs
@@ -1,11 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Windows.Media;
-using System.Windows.Media.Imaging;
 using PixiEditor.Models.DataHolders;
 using PixiEditor.Models.Layers;
 using PixiEditor.Models.Position;
-using PixiEditor.ViewModels;
 
 namespace PixiEditor.Models.Tools.Tools
 {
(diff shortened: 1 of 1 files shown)


PixiEditor/Models/Tools/Tools/FloodFill.cs
```diff
@@ -23,42 +20,41 @@ public override LayerChange[] Use(Layer layer, Coordinates[] coordinates, Color
...
-        public BitmapPixelChanges ForestFire(Layer layer, Coordinates startingCoords, Color newColor)
+        private BitmapPixelChanges ForestFire(Layer layer, Coordinates startingCoords, Color newColor)
         {
             List<Coordinates> changedCoords = new List<Coordinates>();
 
-            Layer clone = layer.Clone();
-            int width = ViewModelMain.Current.BitmapManager.ActiveDocument.Width;
-            int height = ViewModelMain.Current.BitmapManager.ActiveDocument.Height;
+            int width = layer.Width;
+            int height = layer.Height;
```
flabbet, line 28:
```suggestion
            int width = ViewModelMain.Current.BitmapManager.ActiveDocument.Width;
            int height = ViewModelMain.Current.BitmapManager.ActiveDocument.Height;
```

Hey, all awesome, but it doesn't work if you don't set the height and width to `ActiveDocument` due to the nature of dynamic bitmaps. They resize to size of content, so for layer without any pixels, it will be 0x0 and nothing will fill :)

flabbet (changes requested):

PixiEditor/Models/Tools/Tools/FloodFill.cs
brogowski: Oh, didn't know about that - sure, let's change that 👍

flabbet (approved):
Nice, good job

PR #91, opened 2020-11-01
View Model Main refactoring

flabbet:
Cleaned ViewModelMain from the big blob to smaller SubViewModels


flabbet:
Pushed directly to master

PR #96, opened 2020-11-23
Added null check

CPKreuz:
Before checking if the `ActiveDocument` is null it threw a `NullRefrenceException` when checking if the Layer Count is zero


flabbet:
Technically, the user shouldn't be allowed to remove all layers, this is a good check anyway, did you manage to crash the program due to it?

CPKreuz:
When you don't create a new document but left click in the middle it crashes

flabbet (approved):

PR #101, opened 2020-12-03
Added clipboard undo support

### LINKED ISSUE #92: Add Clipboard undo support
Currently, clipboard doesn't have undo support, adding it is quite important.

Trying to undo 'Cut' results in crash

flabbet:
Resolves #92 
Changes:
- Added clipboard undo support
- Added DeletePixels undo support
- Now cut deletes pixels from the selection in the selected layer
- Fixed bug where redo stack didn't reset, while new undo change was added and new undo change was added
- Renamed `StopAction` to `ApplyPreviewLayer`
- Copy now copies selected pixels from the selected layer


CPKreuz (approved):

CPKreuz (changes requested):
I think it's a good idea to fix whatever went wrong

CPKreuz (approved):

PR #107, opened 2020-12-17
Fixed sizeinput empty bug and aspect ratio infinite loop

### LINKED ISSUE #106: Crash when trying to resize the canvas
Trying to resize the canvas causes a stackoverflow exception. To replicate: make a new 256x219 canvas, open resize canvas menu, type 246 in the width field. This only happens when preserve aspect ratio is checked.

flabbet:
Fixed #106 aspect ratio bug, which caused `StackOverflow` with some values.
Also fixed the Behavior which is used in SizeInput to convert numbers to `Num px` format, when focusing out without changing anything + now it supports "{num}px" format, ex. "15px".


flabbet:
@CPKreuz and @Equbuxu review please :)

CPKreuz (approved):
Looks good

Equbuxu:
Yeah, seems to work fine now

Issue #97, opened 2020-11-25
Multiple documents (Tabbed)


CPKreuz:
Currently you can only edit one document at a time, there should be a tab user control where the user can select on what  Document he wants to work on

PR #117, opened 2020-12-19
Tabbed documents support

flabbet:
Solves #97 
Implemented tabbed documents.
Major changes that come with it:
- Changed UndoManager from singleton pattern to instance. Now every document contains it's own undo manager
- Each document have its own selection

Changed DockingManager structure and wrapped ColorPicker into AnchorablePane


CPKreuz (changes requested):
I think you can tell what the issue is

https://user-images.githubusercontent.com/45312141/102700354-edd46700-424c-11eb-99f9-59caefe7ebb0.mp4

Double the document name indicator

doubleTheDocument

flabbet:
Please provide reproduce behavior

CPKreuz (changes requested):
It would be a good idea if you would look why the pipeline is failing before merging

CPKreuz:
Would be nice if you could use tools without selecting the document first

https://user-images.githubusercontent.com/45312141/102717307-c24f8c00-42e1-11eb-9e0a-eff8c261d5b6.mp4

flabbet:
Tbh I don't really have an idea how to fix this, we might add this as a separate issue and try to fix it later

CPKreuz (changes requested):
Happens at around 25 seconds (I was too lazy to cut it). The other document seems to jump sometimes when moving the other one

https://user-images.githubusercontent.com/45312141/102778874-9c82bf80-4393-11eb-952f-b0b4c3503ff1.mp4

CPKreuz:
Document also jumps when zooming in and the document is going outside the area

CPKreuz:
bindingFailures

flabbet:
I can't reproduce it, the way you did on the video. Please write down the steps you did.

CPKreuz (approved):
Can't reproduce anymore, seems to work

PR #121, opened 2020-12-24
Removed tool enum

### LINKED ISSUE #114: Remove ToolType enum
We will need to replace the ToolType enum with something that can be edited at runtime since we want to give extensions the ability to add new custom tools

CPKreuz:
Removed the tools enum

closes #114


PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
```diff
@@ -58,7 +58,7 @@ public void DeletePixels(Layer[] layers, Coordinates[] pixels)
         /// <param name="tool">Tool to execute.</param>
         public void ExecuteTool(Coordinates newPos, List<Coordinates> mouseMove, BitmapOperationTool tool)
         {
-            if (Manager.ActiveDocument != null && tool != null && tool.ToolType != ToolType.None)
+            if (Manager.ActiveDocument != null && tool != null/* && tool.ToolType != ToolType.None*/)
```
flabbet, line 61: why comment

PixiEditor/Models/Tools/Tool.cs
```diff
@@ -9,11 +10,13 @@ namespace PixiEditor.Models.Tools
     public abstract class Tool : NotifyableObject
     {
         private bool isActive;
-        private string actionDisplay = "";
-
-        public abstract ToolType ToolType { get; }
+        private string actionDisplay = string.Empty;
+
+        public string Name => GetType().Name.Replace("Tool", string.Empty);
```
flabbet, line 15: This shouldn't be a getter, instead of `Name =>` it should be `Name = `, because maybe someone would want to change the name in the constructor

PixiEditor/ViewModels/SubViewModels/Main/ToolsViewModel.cs
```diff
@@ -41,12 +42,17 @@ public ToolsViewModel(ViewModelMain owner)
                 new CircleTool(), new RectangleTool(), new EraserTool(), new ColorPickerTool(), new BrightnessTool(),
                 new ZoomTool()
             };
-            SetActiveTool(ToolType.Move);
+            SetActiveTool(typeof(MoveTool));
         }
 
-        public void SetActiveTool(ToolType tool)
+        public void SetActiveTool(Type toolType)
```
flabbet, line 48: This parameter shouldn't be `Type`, instead, it should be a `Tool`, because you don't want anyone to put any type that exists, just a type of tool

PixiEditor/ViewModels/ViewModelMain.cs
```diff
@@ -156,6 +157,12 @@ public bool DocumentIsNotNull(object property)
             return BitmapManager.ActiveDocument != null;
         }
 
+        private Shortcut CreateToolShortcut<T>(Key key, ModifierKeys modifier = ModifierKeys.None)
+            where T : Tool
+        {
+            return new Shortcut(key, ToolsSubViewModel.SelectToolCommand, typeof(T), modifier);
+        }
+
```
flabbet, line 165: This is nice

flabbet (changes requested):
Some minor changes required

Equbuxu (commented):
better than it was before

flabbet (changes requested):
Code looks fine, but the middle mouse button doesn't work

flabbet (approved):
Works good now

PR #122, opened 2020-12-25
Colorpicker update

Equbuxu:
Updated ColorPicker to v.2.0.0, fixed a small crash caused by a missing null check


flabbet (approved):
Beautiful

CPKreuz (approved):
One of the best additions to the project so far imo

PR #123, opened 2020-12-26
Fixed removing logical focus from NumberInput

### LINKED ISSUE #108: Shortcuts disabling themselves randomly
**Describe as detailed as possible when it happens**
Shortcuts don't work randomly.

**Describe what you tried to do in order to fix it**
Always working shortcuts.

flabbet:
Fixes #108

Removed zero-width textbox, which was a hacky fix to remove the focus. Not Ctrl does not get blocked.


flabbet:
@Equbuxu The NumberInput thing is should be fixed now, and Ctrl does work, but please test a little if it still blocks the shortcuts somehow

Equbuxu (approved):
Works great! Love the new textbox behavior. Shortcuts work well now too

flabbet:
Awesome

PR #124, opened 2020-12-27
Added layer merging

### LINKED ISSUE #120: Merge layers

CPKreuz:
Only thing that doesn't work is undoing, I could use some help with that @flabbet 

Closes #120


flabbet:
Please write tests for your changes

PixiEditor/Models/DataHolders/Document.cs
```diff
@@ -425,6 +426,74 @@ public void CenterContent()
...
+            foreach (Layer layer in layers)
+            {
+                Layers.Remove(layer);
+            }
+
+            SetActiveLayer(Layers.IndexOf(finalLayer));
+
+            return finalLayer;
+        }
+
```
flabbet, line 496: `Document.cs` is getting too long, maybe split some methods as other partial one or extension to `Document` class

flabbet (changes requested):
I'll think of a method to implement undo, for now, add tests and change a thing I mentioned.

CPKreuz:
> I'll think of a method to implement undo, for now, add tests and change a thing I mentioned.

Problem is that it crashes now when undoing

flabbet:
I have an idea since we have combined layers, we need to store only BitmapPixelChanges of pixels that are gone due to merging, layer name, offset, width, height, and index, you need to save those pixels during merging

flabbet:
Or what about saving the layers as a .pixi file in the temp folder and load it on demand?

flabbet (changes requested):
I added undo to this, but I found one behavior thing that bothers me. If you merge with below, layers get merged, but in reversed order. I don't think it should work like that

CPKreuz:
> I added undo to this, but I found one behavior thing that bothers me. If you merge with below, layers get merged, but in reversed order. I don't think it should work like that

ah shit, forgot I had to fix that

flabbet (approved):
_    
       | |   
  ___ | | __
 / _ \| |/ /
| (_) |   < 
 \___/|_|\_\

Issue #104, opened 2020-12-05
Undo create and delete layer


flabbet:
Undo should support deleting and creating layers.

PR #136, opened 2021-01-09
Undo remove, add and move layers

flabbet:
Solves #104

Added `StorageBasedUndo` which saves selected layers on disk, and loads them on demand, as part of the Undo and Redo processes. This allows storing layers without memory leak risk. System stores layers in PNG format.

Thanks to the described solution I added Remove layer undo support.

Also wrote tests for `StorageBasedUndo`, but I didn't cover the remove layer, I'll do that tomorrow.


Equbuxu (commented):
Multiple layers are displayed as selected after undoing layer removal. To replicate: make a new canvas, press new layer, right-click base layer, delete it, undo

Equbuxu:
Can't undo adding a layer

CPKreuz (changes requested):
What @Equbuxu said

CPKreuz (commented):
Seems to work

CPKreuz (approved):
Forgot to approve

Equbuxu:
Moving a layer and undoing it prevents you from undoing anything before. To replicate: make a new canvas, draw something, make a new layer, move it below the base layer, attempt to undo everything. You are stuck in an infinite loop of two layers going back and fourth

Equbuxu (changes requested):
In some cases, two layers are selected after undoing layer removal. To replicate: make a new canvas, create 3 new layers, remove the base layer, remove layer 3, remove layer 1, undo three times.

Equbuxu:
Might be outside of the scope of this PR, but changes in layer visibility and opacity should also be undoable

PixiEditor/Models/Undo/StorageBasedChange.cs
```diff
@@ -66,6 +86,23 @@ public Layer[] LoadLayersFromDevice()
...
 
+        public Change ToChange(Action<Layer[], UndoLayer[]> undoProcess, Action<object[]> redoProcess, object[] redoProcessParameters, string description = "")
+        {
+            Action<object[]> finalUndoProcess = _ =>
+            {
+                Layer[] layers = LoadLayersFromDevice();
+                undoProcess(layers, StoredLayers);
+            };
+
+            Action<object[]> fianlRedoProcess = parameters =>
```
Equbuxu, line 116: typo

Equbuxu:
Move layer tool result does not get undone correctly on a layer that was deleted and later restored. To replicate: make a canvas, make a new layer, draw something on it, use the pointer tool to move the drawing somewhere, delete the layer, undo, attempt to undo again.

flabbet:
Okay, fixed the bug. I'll add layer visibility and opacity undo tomorrow

flabbet:
It should work now, I'll added IsVisible and Opacity undo, I need to write and fix tests, but you can review it now

Equbuxu (changes requested):
Alright, I do acknowledge the fact that splitting the review into multiple messages was a stupid idea of mine, but 

this
> Moving a layer and undoing it prevents you from undoing anything before. To replicate: make a new canvas, draw something, make a new layer, move it below the base layer, attempt to undo everything. You are stuck in an infinite loop of two layers going back and fourth

and this

> In some cases, two layers are selected after undoing layer removal. To replicate: make a new canvas, create 3 new layers, remove the base layer, remove layer 3, remove layer 1, undo three times.

still needs fixing

PixiEditor/Models/DataHolders/Document.cs
```diff
@@ -425,6 +471,48 @@ public void CenterContent()
...
+            {
+                Layer layer = Layers.First(x => x.LayerGuid == layerGuid);
+                int index = Layers.IndexOf(layer);
+                bool wasActive = layer.IsActive;
+                Layers.Remove(layer);
+
+                if (wasActive)
+                {
+                    SetNextLayerAsActive(index);
+                }
```
Equbuxu, line 512: Sometimes ActiveLayerIndex in Document ends up with an incorrect value after undo/redo, causing a crash when trying to get ActiveLayer. I believe it is caused by SetNextLayerAsActive not firing when a layer that is not active gets deleted. For example, if there are two layers, and ActiveLayerIndex is 1, deleting layer 0 won't update the index, even though after the deletion there is obviously only 1 layer left and ActiveLayerIndex has to be at 0.
To replicate: make a canvas, make a new layer, delete base layer, undo, select layer 1, redo, attempt to draw.

Equbuxu (changes requested):

flabbet:
Fixed the bug, I will make a separate issue for undoing the selection, this is outside the scope of this PR. @Equbuxu @CPKreuz I would request the review, please

PixiEditor/Models/Controllers/UndoManager.cs
```diff
@@ -35,15 +39,21 @@ public UndoManager(object mainRoot)
...
             lastChangeWasUndo = false;
 
             // Clears RedoStack if last move wasn't redo or undo and if redo stack is greater than 0.
             if (lastChangeWasUndo == false && RedoStack.Count > 0)
-            {
-                RedoStack.Clear();
-            }
+                {
+                    RedoStack.Clear();
+                }
```
CPKreuz, line 56: Fix formatting

CPKreuz (changes requested):

Equbuxu (approved):

PR #142, opened 2021-01-20
Made swatches rounded shadowed rectangles and thin layer borders

flabbet:
Small UI change

image


flabbet:
Replaced shadow with border to keep the style consistent with rest of the app 
image

Equbuxu (approved):

PR #145, opened 2021-01-31
Avalondock theme

flabbet:
- Redesigned AvalonDock with document previews.
- Added Layer previews
- Changed highlighted layer color
- (Hopefully) Improved performance of `CombineLayers`


CPKreuz:
I recommend making sure that the tests are passing

flabbet:
image

CPKreuz (changes requested):
I now looked at it and I really like it

One thing I noticed is when you undock both Color Picker and Layers the docking thing is still there. Would be nice if it would collapse

rightNotGoingAway

CPKreuz:
In other words, why aren't we using the same DockingManager for both document and color picker/layers

CPKreuz:
Also, can you remove the corners between tabs and on the top left
And the close button is still blue

Screenshot (5)

flabbet:
I fixed one CornerRadius below the document tabs, but the rest is, for now, a little too much work. I'll do that eventually. But for now, I would prefer to focus on more important things

PixiEditor/Models/ImageManipulation/BitmapUtils.cs
```diff
@@ -73,6 +80,32 @@ public static WriteableBitmap CombineLayers(Layer[] layers, int width, int heigh
...
+            foreach (var layer in document.Layers.Where(x => x.IsVisible && x.Opacity > 0.8f))
+            {
+                previewBitmap.Blit(
+                    new Rect(layer.OffsetX, layer.OffsetY, layer.Width, layer.Height),
+                    layer.LayerBitmap,
+                    new Rect(0, 0, layer.Width, layer.Height));
+            }
+
+            int width = document.Width >= document.Height ? maxPreviewWidth : document.Width / (document.Height / maxPreviewHeight);
+            int height = document.Height > document.Width ? maxPreviewHeight : document.Height / (document.Width / maxPreviewWidth);
```
CPKreuz, line 104: I get a `DivideByZeroException` for some images, maybe use floats for dividing

CPKreuz (changes requested):
Middle Mouse button should also select the document

CPKreuz (approved):

PR #149, opened 2021-02-11
Implemented pixel-perfect pen option

flabbet:
Added pixel-perfect option to pen tool.


CPKreuz (changes requested):
Love it,

it overwrites already drawn pixels tho (I mean those that in the preview layer)

CPKreuz (approved):
Looks good

PR #150, opened 2021-02-11
Reference layers

CPKreuz:
Template Layers are layers that keep their original size while still being visible on the whole document, they can't be edited and will save the path to the original. Changes made to the original image will automatically be copied to the template layers.

WIP:
Serializing/Deserializing,
use Windows File Picker,
Code cleanup

### DIFF
--- a/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
+++ b/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
@@ -11,7 +11,7 @@ namespace PixiEditor.ViewModels.SubViewModels.Main
 
         public RelayCommand NewLayerCommand { get; set; }
 
-        public RelayCommand NewTemplateLayerCommand { get; set; }
+        public RelayCommand NewReferenceLayerCommand { get; set; }
 
         public RelayCommand DeleteLayerCommand { get; set; }
 
@@ -30,7 +30,7 @@ namespace PixiEditor.ViewModels.SubViewModels.Main
         {
             SetActiveLayerCommand = new RelayCommand(SetActiveLayer);
             NewLayerCommand = new RelayCommand(NewLayer, CanCreateNewLayer);
-            NewTemplateLayerCommand = new RelayCommand(NewTemplateLayer, CanCreateNewLayer);
+            NewReferenceLayerCommand = new RelayCommand(NewReferenceLayer, CanCreateNewLayer);
             DeleteLayerCommand = new RelayCommand(DeleteLayer, CanDeleteLayer);
             MoveToBackCommand = new RelayCommand(MoveLayerToBack, CanMoveToBack);
             MoveToFrontCommand = new RelayCommand(MoveLayerToFront, CanMoveToFront);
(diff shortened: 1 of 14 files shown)


PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -33,6 +39,22 @@ public LayersViewModel(ViewModelMain owner)
             MergeWithBelowCommand = new RelayCommand(MergeWithBelow, CanMergeWithBelow);
         }
 
+        public void NewTemplateLayer(object parameter)
```
flabbet, line 42: NewReferenceLayer

PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -8,6 +11,8 @@ public class LayersViewModel : SubViewModel<ViewModelMain>
 
         public RelayCommand NewLayerCommand { get; set; }
 
+        public RelayCommand NewTemplateLayerCommand { get; set; }
```
flabbet, line 14: NewReferenceLayerCommand

flabbet (changes requested):
It works, however, I am a bit concerned since you need to manually disable features from the Reference layers instead of enabling the ones you need. We might miss that in the future and release the reference layer with things that shouldn't be there. Can you take care of this?

CPKreuz:
> It works, however, I am a bit concerned since you need to manually disable features from the Reference layers instead of enabling the ones you need. We might miss that in the future and release the reference layer with things that shouldn't be there. Can you take care of this? 

This almost looks like I have to rewrite the whole layer system, should I maybe just rewrite the whole project

flabbet:
Well, not really rewrite whole layers system, I guess it would be fine just to work with inheritance. We can find solution to this on discord today in the evening. But if you have idea just write it down here. It might be more work, but much more worth it. It will be much more maintainable.

Equbuxu (changes requested):
Those two "New Layer" and "New Reference Layer" buttons take up way too much space for what they are worth, we should eventually replace them with small pictograms. This PR might be a good place for it.

When moving reference layer using pointer tool it snaps to the "big" pixel grid. This is not a giant deal but it would make more sense for it to be independent.

You can completely break a reference layer by deleting the layer and undoing the deletion. Selecting an area on a reference layer and attempting to drag around it produces a similar result if you drag violently enough. Something weird happens when using line, square, and circle tools on a reference layer.

Document-Clip Canvas seems to take the reference layer into account producing a weird effect, same with Center Content.

Crash: make a new canvas, add a reference layer, delete it, close the canvas tab

Crash: launch PixiEditor, make a canvas, add two reference layers, delete the top one, delete the remaining one. There is about 1/4 chance of a crash, I'm not sure what's causing it, might be the same reason as the previous one.

Crash: try adding a corrupted image as a reference layer

In general, I think reference layers should be better separated from the other code, such that we don't have to check if a given layer is a reference layer everywhere.

CPKreuz:
> In general, I think reference layers should be better separated from the other code, such that we don't have to check if a given layer is a reference layer everywhere.

That's what I was gonna add today

Issue #127, opened 2020-12-29
In some cases, when clicking somewhere while textbox has focus it loses focus too late causing inputted value to be lost or applied to wrong target


Equbuxu:
It's a fairly minor issue but it annoyed me more times than it should've, so I'll document it here. To replicate: have at least two layers present, write a new value in the opacity field, click on the second layer (without clicking anywhere else/removing focus from textbox). The new opacity does not get applied.

Equbuxu:
The best way to fix this (and similar) issues would be to capture mouse clicks when any textbox has focus. This means a click outside any textbox will remove focus from it without triggering anything else, even if there is something clickable under the cursor.

Equbuxu:
Another way is to propagate the updated value on text change and not on focus loss

PR #151, opened 2021-02-12
Small improvements

### LINKED ISSUE #126: Don't restrict canvas to window borders when moving viewport
When the canvas is zoomed out (fits into the window completely) it can't be moved past window borders. While it does make some sense, in practice it is annoying because sometimes it doesn't let you put the canvas in a comfortable position. This issue affects me specifically because my tablet area doesn't cover the whole screen, meaning I have to move canvas sometimes to reach areas even though they are visible.

**Describe the solution you'd like**
Let us move canvas wherever we want

flabbet:
- Fixed ToolSettingColorPicker
- Fixed double click rename layer
- Fixed color picker tool not picking on single click
- Fixed #126 
- Fixed multiple document crash
- Disabled opacity number input when not any document present
- Changed opacity changes on LostFocus instead of PropertyChanged
- Changed assembly version to 0.2.0.0

### DIFF
--- a/PixiEditor/Helpers/Converters/NotNullToBoolConverter.cs
+++ b/PixiEditor/Helpers/Converters/NotNullToBoolConverter.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace PixiEditor.Helpers.Converters
+{
+    [ValueConversion(typeof(object), typeof(bool))]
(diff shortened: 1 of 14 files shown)


Equbuxu (changes requested):
Pipette tool no longer updates the color when dragging the mouse;
Regression of #127: make two layers, type in a new opacity value, select second layer by clicking on the area outside of that layer's name

Equbuxu (approved):

Issue #45, opened 2020-09-12
Create grid lines for canvas


flabbet:
# Overview
There should be an option to enable grid lines, which will show over the canvas, to improve the visibility of individual pixels.

# Implementation idea

User control, which will sit inside the canvas.

# Additional things

- Menu item to enable/disable grid lines
- Shortcut (Maybe Ctrl + ')

yoshivb:
Hi! So I would like to take a look at this one and had a few questions.

**Where should this menu item reside?**
I am not sure if it being under "Document" would be the best.
Maybe adding a new "View" menu and putting it under there? 
Or did you have something else in mind?

**Would a simple approach where the grid is relative to the scale of each pixel on the screen be acceptable?**
As in when you have a 4x4 document the grid would be rather thick and when you have a 50x50 document it would be considerably thinner unless you zoom in.
Example: 4x4 fully zoomed in
![image](https://user-images.githubusercontent.com/9098909/94992127-1458ff00-0588-11eb-9de5-8ba09fbdb845.png)

Example: 50x50 zoomed to take up roughly the same size on screen
![image](https://user-images.githubusercontent.com/9098909/94992181-7e71a400-0588-11eb-8a04-48edc32c4952.png)

flabbet:
Yes, the `View` menu is a good place to store the setting, What do you mean under Document? My initial idea was something like that
```xaml
<PixelGridLines>
    <Canvas/>
</PixelGridLines>
```

Where `<PixelGirdLines>` is the control that overlays lines on its child. This is just a thought, you can do it another way.

Yes, the thickness of lines should be consistent. For example here is 4x4 zoomed-in other graphics software
![image](https://user-images.githubusercontent.com/25402427/94992420-3bb0cb80-058a-11eb-81b1-5abce5a6cabc.png)

And here is 50x50
![image](https://user-images.githubusercontent.com/25402427/94992430-48352400-058a-11eb-9d35-5a8dee5a29b7.png)

We want to achieve a similar effect. As you can see on 50x50 there are "groups" of lines, 5x5 big squares with 10x10 individual pixels inside. It should be dependent on the zoom percentage. As far as I remember ZoomPercentage in `ViewModelMain`  stores only changes done by `ZoomTool`, so you would need to update it in `<MainDrawingPanel>` control but I might be wrong.

yoshivb:
The first question was regarding the enable/disable menu item.
Whether it should be under "Document" (like shown below) or possibly a new menu?
[...]

PR #154, opened 2021-02-15
Added grid lines

flabbet:
Solves #45.

Toggle grid lines available under Views Menu
Shortcut for toggling: Ctrl + `

### DIFF
--- a/PixiEditor/Helpers/Converters/IntToViewportRectConverter.cs
+++ b/PixiEditor/Helpers/Converters/IntToViewportRectConverter.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Globalization;
+using System.Windows;
+using System.Windows.Data;
+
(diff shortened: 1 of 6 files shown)


CPKreuz (changes requested):
Can you make the lines dashed? And the dashed ones in a different color.
They should also be in front of preview layers

flabbet:
Dashed like an option or instead of solid ones?

flabbet:
Fixed

CPKreuz (changes requested):
Only 10% of the canvas is dashed now?

weridDashButOk

CPKreuz (changes requested):
Can you also add a line that marks the middle

flabbet:
This would be a part of guides.
And not all canvas sizes have a middle line. So it's outside of the scope of this PR

PR #155, opened 2021-02-22
Multiple layers selection

flabbet:
This PR adds a lot of stuff along with multiple layers selection. 

Selecting multiple layers can be done in 2 ways:
CTRL + Left click to toggle individual
SHIFT + Left click to select range

Stuff that supports multiple layers:
- Move tool (moving all selected layers)
- Merge selected (Context menu)
- Delete (context menu)
- Copy to clipboard 
- Remove pixels (shortcut: delete key).

Also changed color of the dropdown, so it has more contrast with the background on the toolbar

Still, tests TODO, so review the PR please, suggest maybe what else should support multiple layers, and find bugs, while I am writing tests.

Update: Wrote some tests


CPKreuz (changes requested):
Merge with above/below only merges with one layer of the selection, maybe give that layer an extra color

flabbet:
Disabled Merge with above/below for multiple layers selected

CPKreuz (changes requested):
How about being able to edit the opacity of multiple layers

flabbet:
@Equbuxu What do you think about this?

Equbuxu (changes requested):
Have two layers with the first one of them selected, hold down ctrl, select the second layer, deselect the first layer, release ctrl, try drawing something. Result: the second layer is selected, but  the brushstrokes are saved on the first one. Clip Studio Paint solves this problem by using two different types of selection: https://prnt.sc/10de0bx. One selection is highlighted and signifies the main layer, it can only be changed by left-clicking a layer (and consequently deselecting everything)

Crash: Make a new 16x16 file, make a second layer, draw a diagonal line going from top left corner to bottom right corner on the base layer, draw a perpendicular line going from the top right corner to bottom left corner on the second layer. Now, select both layers, choose area selection tool, and select the 2x2 square of pixels in the middle of the screen. Choose pointer tool and drag the selected area to the right outside the canvas. Release the left mouse button there, then move the area back and make sure to position it on top of other leftover pixels. Release left mouse button again, then grab the area and drag it to the right again. Release left mouse button. These steps are very specific, but the bug is actually fairly easy to trigger and that's just one of the ways to do so.

Crash: make a new canvas, press new layer, hold down ctrl, click new layer 1 to deselect it, release ctrl, right click new layer 1 and select delete, attempt to change opacity without touching anything else.

Crash similar to previous: make a new canvas, make a second layer, ctrl-deselect and delete second layer, try to draw something

CPKreuz:
> One selection is highlighted and signifies the main layer,

I wanted to say the same thing but didn't know how to say it

PixiEditor/Models/Layers/Layer.cs
```diff
@@ -58,6 +59,16 @@ public Layer(string name, WriteableBitmap layerBitmap)
 
         public Dictionary<Coordinates, Color> LastRelativeCoordinates { get; set; }
 
+        public string LayerHighlightColor
+        {
+            get => IsActive ? layerHighlightColor : "#00000000";
+            set
+            {
+                layerHighlightColor = value;
+                RaisePropertyChanged(nameof(LayerHighlightColor));
```
CPKreuz, line 68: You could use the new SetProperty Method that I added in undo selection here

```suggestion
                SetProperty(ref layerHighlightColor, value);
```

CPKreuz (changes requested):

CPKreuz (changes requested):
Select multiple layers, merge selected, undo, layers now have a weird background color

CPKreuz (changes requested):
Right click does not select the layer, but opens the context menu for it

flabbet:
> Right click does not select the layer, but opens the context menu for it

This is by design, it's not a bug

flabbet:
> Select multiple layers, merge selected, undo, layers now have a weird background color

Can't reproduce, provide .pixi file and which layers are you talking about

flabbet:
Fixed

Equbuxu (changes requested):
Make a new canvas, make a new layer, select base layer, click on the down button of New Layer 1, draw something. Result: currently selected layer is Base Layer but you are drawing on New Layer 1. Shift-click New Layer 1. Now there is no "strongly" selected layer.

flabbet:
Fixed

CPKreuz (approved):
Looks good

flabbet:
Hot update: Made `CenterContent` selection dependend, now it only centers selected and visible layers.

Equbuxu (changes requested):
[I didn't want to write a long description again.](https://www.youtube.com/watch?v=LGdHy-tffbY)

PR #159, opened 2021-03-07
Undo selection

CPKreuz:

### DIFF
--- a/PixiEditor/Helpers/SelectionHelpers.cs
+++ b/PixiEditor/Helpers/SelectionHelpers.cs
@@ -6,30 +6,38 @@ using PixiEditor.Models.Undo;
 
 namespace PixiEditor.Helpers
 {
-    public class SelectionHelpers
+    public static class SelectionHelpers
     {
-        public static void SelectionSet(object[] arguments)
-        {
-            Document document = (Document)arguments[0];
-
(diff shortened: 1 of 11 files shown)


PixiEditor/Helpers/SelectionHelpers.cs
```diff
@@ -0,0 +1,35 @@
...
+using PixiEditor.Models.DataHolders;
+using PixiEditor.Models.Enums;
+using PixiEditor.Models.Position;
+using PixiEditor.Models.Undo;
+
+namespace PixiEditor.Helpers
+{
+    public class SelectionHelpers
+    {
+        public static void SelectionSet(object[] arguments)
```
flabbet, line 11: Change the name to SetSelectionProcess, generally keep the names for undo methods using OriginalName + Process

flabbet (changes requested):
Perfect, I love how the code is clean. One thing I would change is not to deselect on undo. Only remove selection when user wants to (deselect manually or via Undo). You didn't add this, it was before, but because we have undo now, you can remove that behavior.

flabbet (changes requested):
Adding and subtracting selection should be undoable

CPKreuz:
> Adding and subtracting selection should be undoable

Well it is, it's just treated as a new one, will fix

flabbet (approved):
Fabulous

PR #160, opened 2021-03-08
Added Debugging Tools

CPKreuz:
Not sure if we need this

### DIFF
--- a/PixiEditor/ViewModels/SubViewModels/Main/DebugViewModel.cs
+++ b/PixiEditor/ViewModels/SubViewModels/Main/DebugViewModel.cs
@@ -1,0 +1,23 @@
+﻿using System.Diagnostics;
+using PixiEditor.Helpers;
+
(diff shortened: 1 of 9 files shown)


PixiEditor/ViewModels/SubViewModels/Main/DebugViewModel.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public RelayCommand BreakCommand { get; set; }
+
+        public DebugViewModel(ViewModelMain owner)
+            : base(owner)
+        {
+            BreakCommand = new RelayCommand(Break, CanBreak);
+        }
+
+        public static void Break(object parameter)
```
flabbet, line 16: I don't see the point of this menu option, while we can keep this VM, I would remove this Command

flabbet (changes requested):

flabbet (approved):

PR #162, opened 2021-03-11
Added canvas context menu and duplicate layer command

CPKreuz:
Added a canvas context menu, containing (De)select all, Copy, Cut, Paste, Bake Pizza and added a duplicate layer command to the layer list

### DIFF
--- a/PixiEditor/Models/DataHolders/Document/Document.Layers.cs
+++ b/PixiEditor/Models/DataHolders/Document/Document.Layers.cs
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Windows;
 using System.Windows.Media.Imaging;
 using GalaSoft.MvvmLight.Messaging;
+using PixiEditor.Helpers;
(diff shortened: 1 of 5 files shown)


flabbet (changes requested):
Insert (1) at the end of the duplicated layer name, add undo support. I don't think you need to write any undo process, you can reuse what already exists

PixiEditor/Models/DataHolders/Document/Document.Layers.cs
```diff
@@ -437,6 +469,65 @@ private void RemoveLayerProcess(object[] parameters)
...
+                        if (highestValue == null)
+                        {
+                            highestValue = 0;
+                        }
+                    }
+                }
+            }
+
+            return highestValue;
+        }
```
flabbet, line 529: This is a bit too long function, split it a little, please :)

flabbet (changes requested):
The code is clean, everything works, I was unable to break it, you've reused the code well and even wrote a test for suffixes, perfect. The only thing I need to ask you is to write a test for the Duplicate layer. Nothing fancy, just if it's added to Layers and is active, you don't need to compare contents or test undo.

flabbet (approved):
👌

PR #164, opened 2021-03-20
Added Shortcut and Startup Popup

CPKreuz:
Shortcut window shows all the shortcuts in `ShortcutManager`

Startup Window (aka the HelloThereWindow) is the new window that appears when you open PixiEditor (can be disabled as previously)

It contains the name of this wonderful raster graphics pixel editor called "PixiEditor", the version, buttons for opening and creating documents, a list of recently opened documents and three buttons for an invite to our wonderful discord server (wich you can join here https://discord.gg/tzkQFDkqQS (why tf am I writing this in the pull request description)) an link to our subreddit (https://reddit.com/r/PixiEditor) and a link to this site (https://github.com/PixiEditor/PixiEditor)


CPKreuz:
I still wanna add that the wrap panel makes an actually nice grid

flabbet (changes requested):
Let's start with things that are visible as users, after that let's focus on code.

So first: Make recent files even (spacing, width, height), too long names should be replaced with an ellipsis (For example from `Baby yoda new.pixi` to `Baby yoda new...`, so all names have the same max width).

Second: Add to views menu ability to open startup window

PixiEditor/Models/ImageManipulation/BitmapUtils.cs
```diff
@@ -106,6 +107,25 @@ public static WriteableBitmap GeneratePreviewBitmap(Document document, int maxPr
             return previewBitmap.Resize(width, height, WriteableBitmapExtensions.Interpolation.NearestNeighbor);
         }
 
+        public static WriteableBitmap GeneratePreviewBitmap(IEnumerable<SerializableLayer> layers, int width, int height, int maxPreviewWidth, int maxPreviewHeight)
```
flabbet, line 110: What did you change in this function?

flabbet (changes requested):
Add shortcut on F1 to open Shortcuts window

PixiEditor/Models/ImageManipulation/BitmapUtils.cs
CPKreuz: Made it use `SerializableLayer` instead of `Document`

flabbet: You can reuse a lot of code. DRY

flabbet (approved):

flabbet:
Next time write more tests

PR #172, opened 2021-04-07
Group layers

flabbet:
This PR is ready to review, I am still waiting for PixiParser changes, so this will be added later.

Added possibility to create layer groups with nesting. The whole system is done with drag and drop. Also changed the look of layers. Added also raw layers window, for debugging purposes.
Also fixed shortcuts not firing bug (hopefully) and changed checker background.

Todo list:

- [x] Drag n drop sometimes drags somewhere else bug
- [x] Change group name
- [x] Delete group
- [x] Prevent from moving folder inside its children
- [x] Selected group visuals
- [x] Multiple layers inside group delete and merge bug
- [x] Insert layers relative to selected layers
- [x] Full space layers and groups control
- [x] New layer, a new group, delete a layer, delete group pictograms
- [x]  Group preview
- [x] Undo (merge, add layers, remove a layer, move layers, move groups)
- [x] Fixed even straight line for bigger thickness lines
- [x] Opacity undo
- [x] Move item outside group (drag n drop)
- [x] PixiParser support
- [x] Tests


PixiEditor/Properties/AssemblyInfo.cs
```diff
@@ -10,7 +10,7 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("PixiEditor")]
 [assembly: AssemblyProduct("PixiEditor")]
-[assembly: AssemblyCopyright("Copyright PixiEditor © 2018 - 2021")]
+[assembly: AssemblyCopyright("Copyright PixiEditor © 2017 - 2021")]
```
Equbuxu, line 13: do we really need years here

Equbuxu (changes requested):
make a new canvas, make a new layer. The delete layer button is inactive until you click on the created layer, even though it is selected already. 

Now close the canvas. For some reason making a new layer is not considered a change, so the canvas closes without prompting to save changes.

Regardless of that, now there are no opened documents. Yet, the layers panel still shows the layers, and selecting one of them causes a crash.

The horizontal scrollbar in the layers panel is kinda tiny and doesn't move all the way to the sides (only visible if you make enough nested layers)

The visibility of layers inside a group should not be directly tied to the visibility of said groups. In other words, if you hide and unhide the group which contains already hidden layers, those layers must preserve their hidden state. Ideally, groups should also have their own opacity that is applied on top of the opacity of all the layers inside.
Related bug: make two nested groups, hide both of them, make one of them visible again. The layer inside becomes visible, even though the second group is still hidden. The same happens if you unhide the layer directly.

The fact that a group can't be empty is questionable for UX.

PixiEditor/Properties/AssemblyInfo.cs
flabbet: no

CPKreuz (changes requested):
When double clicking a group name to rename it, it closes the group

Can you add that when you select multiple layers and hit the new layer group button it puts all the selected layers in the new group

Group opacity is not applied in navigation window

I also found that sometimes when I wanted to click a layer I accidently moved it

Equbuxu (changes requested):
**The navigator window ignores group visibility**
Draw something, put the current layer in a new group, hide the group. The layer is still visible on the navigator.

**Non-intuitive behavior when creating a new layer while the selected group is collapsed**
Right now pressing a new layer button seemingly does nothing because the new layer is not visible. The group should either:
- Be expanded to show the new layer or 
- The new layer should be created above the collapsed group (imo the preferred option)
In addition to that, right now a new layer is generally created below the selected one, and it would make a lot more sense to put it above since that's where you generally want it to be.
Furthermore, the current group should generally be automatically expanded when things happen inside of it. For example, if the currently selected layer is inside of a hidden group (which itself is not selected) pressing the new layer button should expand the group (and create a new layer inside). The same goes for new groups, they should be expanded by default.

**Dragging a layer onto a group should put it inside that group**
^

**Invalid opacity after rearranging groups**
Make this setup:

Put the top group inside the bottom one:

Expected result: Both layers should be translucent
Actual result: New Layer 1 is fully opaque

**Invalid visibility after putting layer inside hidden group**
Hide a group, put a visible layer inside of it. The layer should no longer be visible. The same goes the other way around. Put a layer inside a group, hide the group, drag the layer outside that group. The layer is still hidden but should be visible.

**It's not possible to delete entire groups at once**
Pressing the delete button while a group is selected should delete the entire group, especially if the group is collapsed.

**Opacity is applied to a wrong layer**
Make a new canvas, draw something on the base layer. Press "New Layer". Change Opacity to 50%. The opacity of the base layer gets changed even though the new layer is selected.

**Group preview is completely invisible if the only layer inside has an opacity that is <100**
Not sure if this is intended but seems weird to me

**The base layer that's created when you delete the existing one has no boundaries**
Make a new canvas, press the delete layer button. You can now draw anywhere below and to the right of the canvas. This bug probably existed before, but it wasn't possible to delete the base layer.

Equbuxu (changes requested):
**Double-clicking eye button causes another layer to activate name editing field**
You need a group with one layer. First, click the group, then, double-click the eye button of the layer. PixiEditor thinks that you wanted to change the group's name. I think double-clicking on the eye button should not activate the name editing field at all, even that of the same layer.

**You can't change the opacity of a layer if it's inside a group**
Actually, sometimes you can, but most of the time the opacity field does nothing

**Binding errors**
Not sure if anything can be done there, just pointing out that there are some new ones.

### Stuff from the previous review that wasn't fixed
**Creating a new group inside of a collapsed group does not expand the collapsed group**
Creating a new layer inside does, same should work for groups

Equbuxu (changes requested):
**Weird crash when dragging selected text**
Double click a layer or a group to rename it. Now try to drag the selected text. `Dispatcher processing has been suspended, but messages are still being processed` in `LayerStructureItemContainer_MouseMove` inside the loop.

**Crash in CalculateTopIndex**
That's the second time I got it, still can't figure out how to replicate it, but here is the [crash log](https://pastebin.com/kuEgmP6Q). The general idea is: move a lot of stuff around creating and deleting layers and groups, then make a few nested groups with a few layers on each nesting level, move a bit more stuff around, start deleting layers. 

### UX

**Group is expanded for no reason**
Make a new canvas, make a new group, select that group and hide it. Now, make a new layer. As expected, the layer is created above the group, but the group is also expanded for no reason.

**Inconsistent layer creation behavior**
Make a new canvas, press `New Layer`, select _"Base Layer"_, press `New Group`, collapse the group, press `New Layer`. The layer is created inside the group even though the group is collapsed and it was supposed to be created above.

**A group is not expanded when it's supposed to**
Delete a layer that is inside of a collapsed group (the group should have other layers too). The group is not expanded.

Equbuxu (commented):
That's not even really a proper review but something that I found. Overall I don't see any terrible bugs, but will test more

## UX
**"Inconsistent layer creation behavior" from the previous review isn't fixed**

**The layers window has a weird white square in the bottom right corner**
It only appears when both scrollbars are visible

CPKreuz (changes requested):
The delete layer button is clickable when there's no layer

CPKreuz:
I didn't pull, sorry

CPKreuz (commented):
I got a idea for a big feature

What if the close button would be red?

CPKreuz (changes requested):
Layer Setup:
layerSetup

When you select the New Layer 2 and then shift select the Base Layer, all layers are selected but not the group

selected

flabbet:
It's just a visual thing, let's ignore it for now

PR #176, opened 2021-04-12
Added commands to open local app data, roaming app data and install location

CPKreuz:


flabbet (changes requested):
Tbh, I think it should be as a button in the settings window. It's a useful option not only while debugging

flabbet (approved):

PR #185, opened 2021-04-30
Added Microsoft Store badge to readme in installation section

flabbet:


CPKreuz (changes requested):
Maybe make the Or bold

Equbuxu (commented):
Since you are changing the readme, what do you think about

- Adding a "Contents" section at the top
- Removing the "Gallery" section and moving the link from it to "Featured content"
- Removing the "FAQ" section and moving the FAQ link to "Support"

The badge is looking good

Equbuxu (approved):
I did not realize GitHub has autogenerated contents, that's great

Issue #178, opened 2021-04-17
Fix max main window height


CPKreuz:
Currently the max window height is set to `SystemParameters.MaximizedPrimaryScreenHeight` [(see here)](https://github.com/PixiEditor/PixiEditor/blob/master/PixiEditor/Views/MainWindow.xaml.cs#L34)

This works fine as long as the user doesn't put PixiEditor on a monitor with a different resolution or changes the screen resolution while PixiEditor is running

Now, why is the property set in the first place. That's because we're using our own top bar (where the close button and file, select, ... buttons are located), to hide the default windows one we're setting `WindowStyle="None"` this has the side effect that PixiEditor goes over the task bar wich is fixed by setting `MaxHeight`

flabbet:
@CPKreuz Did you fix it in #173?

Equbuxu:
https://stackoverflow.com/questions/20941443/properly-maximizing-wpf-window-with-windowstyle-none
this should help

PR #190, opened 2021-05-30
Fix max height bug

CPKreuz:
Fixed bug from #178 thanks to @Equbuxu


flabbet:
I find personally very frustrating when close, minimize and change window state buttons are not perfectly aligned to the window border, since I usually yeet my cursor to max top-right and just click to close.

CPKreuz:
> I find personally very frustrating when close, minimize and change window state buttons are not perfectly aligned to the window border, since I usually yeet my cursor to max top-right and just click to close.

May I introduce you to Alt+F4

(Also: Will fix)

flabbet (approved):
Seems about right

PR #194, opened 2021-06-11
Tool Dependency Injection

CPKreuz:


CPKreuz:
Tests will be fixed soon

PixiEditor/Models/Tools/Tools/FloodFill.cs
```diff
@@ -1,14 +1,16 @@
...
 using PixiEditor.Models.Position;
-using PixiEditor.ViewModels;
 
 namespace PixiEditor.Models.Tools.Tools
 {
     public class FloodFill : BitmapOperationTool
     {
+        public BitmapManager BitmapManager { get; set; }
+
         public FloodFill()
```
flabbet, line 14: I think that dependencies should be passed through the constructor, not using property modification using binding injection, this works fine, but for someone who doesn't know how DI works and just want to create a tool, this might be confusing (how is this thing assigned? Etc.), but when he sees in examples that there is a constructor parameter, this magic disappears, everything is pretty clear. I am open to discussion, I don't request changes. But imo this should work like that. Afaik ASP.NET does that

CPKreuz: I can confirm that ASP.Net does it the way you described it [(Example here)
](https://docs.microsoft.com/de-de/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#code-try-5)
Now the library that PixiBot uses to communicate with Discord does it the way that I did here [(Example here)](https://github.com/discord-net/Discord.Net/blob/dev/samples/02_commands_framework/Modules/PublicModule.cs#L13)

---

I would say that both methods should be easy for beginners and I wouldn't say there's a big difference

If they just read the docs, they will probably just read something like this

> The assignment of properties like BitmapManager is handled automatically by the ToolBuilder.


> The arguments for the constructor (like BitmapManager) are passed automatically by the ToolBuilder.

---

I like both ways and both are an improvement to what we have atm

I'd be interested in what @Equbuxu says to this

flabbet: Let's do it .NET way, passed through constructor

flabbet (changes requested):
comment above

flabbet (approved):
gj

Issue #188, opened 2021-05-26
MSIX still checks updates on startup


flabbet:
Basically title. Whenever there is an update, MSIX version still downloads it and displays popup

PR #199, opened 2021-06-30
Improved touch and stylus support

CPKreuz:


flabbet:
So this PR is without reference layer right?

CPKreuz:
Nope, but I will force push in a bit

flabbet:
It feels a little laggy to use

flabbet (changes requested):
Touch gestures work, new file animation is nice, but drawing using touch screen doesn't work

flabbet (changes requested):
When opening a document, you can't zoom using touch, you need to select any tool, event selected one, then it works

flabbet (approved):
Everything looks good

CPKreuz:
This should also fix #188

PR #202, opened 2021-07-02
Reference layer

flabbet:
Added reference layer in LayersManager window


CPKreuz (changes requested):
1. You can select a preview layer even when there's no document opened

2. When you open a second document it shows that it has a reference layer even though it shouldn't

3. I think Add reference layer sounds better

CPKreuz (changes requested):
Dragging Layers doesn't work

flabbet:
Fixed

CPKreuz (commented):
Just saw this in the last commit, use nameof(Property) instead of a string for properties and maybe make the `LayerGroupControlDataName` and `LayerContainerDataName` static and set the value to `typeof(LayerGroupControl).FullName`

flabbet:
Oh, I didn't know about `typeof(x).FullName` thing

CPKreuz (approved):

Equbuxu (commented):
Fix warnings

Equbuxu (changes requested):
wrong type of review

Equbuxu (approved):

Issue #200, opened 2021-07-02
Opening a file with a long name crashes PixiEditor


Equbuxu:
DiscordViewMode.cs -> UpdatePresence -> richPresence.Detail = ...

DiscordRPC.Exceptions.StringOutOfRangeException: 'Length of string is out of range. Expected a value with a maximum length of 128'

Issue #197, opened 2021-06-28
ActionDisplay of tools doesn't get updated


Equbuxu:
Each tool has its own `ActionDisplay` which sometimes gets updated from inside the tool. For example, the default ActionDisplay of MoveTool is _"Hold mouse to move selected pixels. Hold Ctrl to move all layers."_ but if you hold down Ctrl it changes to _"Hold mouse to move all selected layers."_. 

However, the modified ActionDisplay doesn't get propagated into the view. This happens because the `TextBlock` in **`MainWindow.xaml`** is bound to `ActionDisplay` property of **`ViewModelMain.cs`** instead of `ActionDisplay` property of the current tool. `ActionDisplay` in **`ViewModelMain.cs`** simply returns the `ActionDisplay` property of current tool, but when `ActionDisplay` of the current tool gets updated `RaisePropertyChanged` is not called on **`ViewModelMain.cs`**'s `ActionDisplay` thus the view is not updated.

PR #204, opened 2021-07-08
Implemented Magic Wand Tool

### LINKED ISSUE #193: Magic Wand tool
Selects all pixels that have the same color. It should have those options:
- Selection Modes: Normal, Add, Subtract, Intersect
- Global/Continous. Global looks through the whole image, Continuous looks through connected pixels (using flood fill or similar)
- Sampling: Layer/Image
- Tolerance: How close a color should be to the clicked color to be selected

CPKreuz:
Closes #193, #197 and #200


Equbuxu (dismissed):
zoom tool actiondisplay doesn't work (pressing down alt doesn't change the hint)
make a 9999*9999 canvas -> click with the magic wand -> pixieditor hangs
please implement other things from the magic wand issue

also maybe don't implement them just yet because the selection is likely to be rewritten together with the move tool and I think flabbet was going to work on that soon

CPKreuz:
> zoom tool actiondisplay doesn't work (pressing down alt doesn't change the hint)
This actually appears to be because windows is selecting the menu bar

> make a 9999*9999 canvas -> click with the magic wand -> pixieditor hangs
I guess that's just an general flood fill issue

flabbet (changes requested):
One feature I would definitely add, is multi-layer support. Tolerance would be good too, but it's all right if it's not here, since we are working with pixel-art images.

PixiEditor/Models/ImageManipulation/BitmapUtils.cs
```diff
@@ -40,16 +40,16 @@ public static WriteableBitmap BytesToWriteableBitmap(int currentBitmapWidth, int
...
             WriteableBitmap finalBitmap = BitmapFactory.New(width, height);
 
             using (finalBitmap.GetBitmapContext())
             {
-                for (int i = 0; i < layers.Length; i++)
+                for (int i = 0; i < layers.Count(); i++)
                 {
-                    float layerOpacity = structure == null ? layers[i].Opacity : LayerStructureUtils.GetFinalLayerOpacity(layers[i], structure);
-                    Layer layer = layers[i];
+                    float layerOpacity = structure == null ? layers.ElementAt(i).Opacity : LayerStructureUtils.GetFinalLayerOpacity(layers.ElementAt(i), structure);
```
flabbet, line 51: `ElementAt(i)` can be called once and saved into variable, it's more efficient.

PixiEditor/Models/Tools/Tools/MagicWandTool.cs
```diff
@@ -0,0 +1,79 @@
...
+            Document document = BitmapManager.ActiveDocument;
+            Layer layer;
+
+            if (documentScope == DocumentScope.SingleLayer)
+            {
+                layer = BitmapManager.ActiveLayer;
+            }
+            else
+            {
+                layer = new Layer("_CombinedLayers", BitmapUtils.CombineLayers(document.Width, document.Height, document.Layers, document.LayerStructure));
```
flabbet, line 48: Implement caching, cache should be saved until tool is changed.

flabbet (changes requested):

PixiEditor/Models/Tools/Tools/MagicWandTool.cs
CPKreuz: What if the document/canvas get's resized?

flabbet: Well of course, you need to consider this kind of scenarios

flabbet: In that case, cache should be refreshed

flabbet (approved):
flabbet certified

flabbet (changes requested):
I was too fast, the moment you shutdown PixiEditor, it crashes
image

flabbet (changes requested):
Fix tests :p

flabbet (approved):

PR #211, opened 2021-07-26
Skia Rework

flabbet:
~~Should work faster than forest fire. This is a preparation for direct bitmap rework~~

PixiEditor used to do all image processing using WriteableBitmap/WriteableBitmapEx. That approach proved to have multiple problems such as:
- All drawing is done on the cpu which is too slow.
- WriteableBitmapEx requires WriteableBitmaps with premultiplied alpha. Alpha premultiplication effectively limits color resolution for translucent pixels.
- WriteableBitmap being a part of WPF is not cross platform.
- Advanced features like layer masks or blending modes require hardware acceleration and can't be feasibly implemented with WriteableBitmaps.

This PR changes the entirety of PixiEditor to use SkiaSharp library for drawing. It also changes all tools to work with the bitmaps directly using SkiaSharp drawing functions wherever possible. Apart from that there are some minor changes, mainly the optimization of fill algorithm.


Equbuxu (changes requested):
Make a 512x512 canvas, draw a line diagonally going from top left to bottom right (the line should separate the canvas in two). Using the fill tool, double click on one of the halves. Wait for it to process and then move your mouse to the second half. The second half gets filled too because the mouse button got stuck (doesn't happen all the time, try again if it didn't).

I did some not-so-scientific testing to find out that the new algorithm is about 6.5 times faster than the previous one. (474 ms vs 72 ms on the image I tested). Howerer, the previous algorithm used to run in 0 ms in cases where  you are trying to fill the area with it's same color (e.g. fill a white area with white). The new algorithm takes about the same time in this case (72 ms). The fill tool gets used at least twice every time you click, and way more than twice if you click and drag. Because of that the new algorithm can actually be slower, as the previous one only ran for a long time once while the new one takes the same time to run many times in a row.

TLDR: make it so fill tool is only called once per click and/or don't fill the area if it's color matches the fill color.

Xiphereal:
Hi, I was wondering if I could lend a hand. Several questions:

1. Is this "Skia Rework" the conversion mentioned at #221?
2. What tasks are left for this PR? Anything I can help with?

Equbuxu:
> Hi, I was wondering if I could lend a hand. Several questions:
> 
> 1. Is this "Skia Rework" the conversion mentioned at [Convert to SkiaSharp #221](https://github.com/PixiEditor/PixiEditor/issues/221)?
> 2. What tasks are left for this PR? Anything I can help with?

Hello!

1. Yes
2. [Here is the list of tasks](https://github.com/PixiEditor/PixiEditor/projects/7)

I should mention however that this PR introduces a lot of small changes all over the codebase, so you will need a decent understanding of the code before you can really help. In other words none of the tasks here are "good first issues". I've updated the first message in the PR, you can check it out for more context. You should also join our [discord server](https://discord.gg/qSRMYmq) if you wanna contribute, that's where most of the discussions take place.

Xiphereal:
I've checked out the origin/quickfill branch, but I'm having several missing types when trying to compile:

untitled

May be some unpushed changes (`LayerChange`and `PixelChangesController`) from your local working directory?

Equbuxu:
> I've checked out the origin/quickfill branch, but I'm having several missing types when trying to compile:
> 
> untitled
> 
> May be some unpushed changes (`LayerChange`and `PixelChangesController`) from your local working directory?

It looks like some tests got broken after we've merged direct bitmap branch into quickfill. I actually didn't realize that before, that's why I didn't mention it in #227 . Regardless of that it's too early to fix them because the code they test is currently being rewritten

flabbet:
Yeah. Ignore the tests for now. You can compile PixiEditor itself, without PixiEditorTests

Xiphereal:
> 
> 
> Yeah. Ignore the tests for now. You can compile PixiEditor itself, without PixiEditorTests

Then I think I'll be looking for other opened tasks, non-dependant of this Skia Rework :) Thanks for the info.

Xiphereal:
I read that the `Bitmap` work will be replaced by Skia operation and types. Is [this](https://github.com/PixiEditor/PixiEditor/blob/0034798790eda65099b6276faf9a3fa0a01ed242/PixiEditor/Models/ImageManipulation/BitmapUtils.cs#L78) also the case?

flabbet:
It is already replaced [here](https://github.com/PixiEditor/PixiEditor/blob/0034798790eda65099b6276faf9a3fa0a01ed242/PixiEditor/Models/ImageManipulation/BitmapUtils.cs#L144-L161)

Xiphereal:
> 
> 
> It is already replaced [here](https://github.com/PixiEditor/PixiEditor/blob/0034798790eda65099b6276faf9a3fa0a01ed242/PixiEditor/Models/ImageManipulation/BitmapUtils.cs#L144-L161)

But the return type is `WriteableBitmap`, which presumably should be replaced, right?

> 
> 
> ~Should work faster than forest fire. This is a preparation for direct bitmap rework~
> 
> PixiEditor used to do all image processing using WriteableBitmap/WriteableBitmapEx. That approach proved to have multiple problems such as:
> 
>     * All drawing is done on the cpu which is too slow.
> 
>     * WriteableBitmapEx requires WriteableBitmaps with premultiplied alpha. Alpha premultiplication effectively limits color resolution for translucent pixels.
> 
>     * WriteableBitmap being a part of WPF is not cross platform.
> 
>     * Advanced features like layer masks or blending modes require hardware acceleration and can't be feasibly implemented with WriteableBitmaps.
> 
> 
> This PR changes the entirety of PixiEditor to use SkiaSharp library for drawing. It also changes all tools to work with the bitmaps directly using SkiaSharp drawing functions wherever possible. Apart from that there are some minor changes, mainly the optimization of fill algorithm.

PR #212, opened 2021-07-26
BitmapOperationsUtility :  Shift functionality should not apply to PenTool

### LINKED ISSUE #203: Pen Tool's internal Line Tool changes behavior when shift is held
The Pen Tool's internal Line Tool tries drawing even lines when shift is held.

rebizu:
BitmapOperationsUtility did not take in to account if the tool used was PenTool.
This patch adds a check that makes it so the PenTool is excluded from the "shift" functionality 

This fixes #203

### DIFF
--- a/PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
+++ b/PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
@@ -128,8 +128,8 @@ namespace PixiEditor.Models.Controllers
             int thickness = sizeSetting != null ? sizeSetting.Value : 1;
 
             bool shiftDown = Keyboard.IsKeyDown(Key.LeftShift);
-            var isShapeTool = tool is ShapeTool;
-            if (shiftDown && !isShapeTool)
+           
+            if (shiftDown && tool.UsesShift)
             {
                 bool mouseInLine = MouseCordsNotInLine(mouseMoveCords, thickness);
 
--- a/PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
+++ b/PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
@@ -128,8 +128,8 @@ namespace PixiEditor.Models.Controllers
             int thickness = sizeSetting != null ? sizeSetting.Value : 1;
 
             bool shiftDown = Keyboard.IsKeyDown(Key.LeftShift);
-
-            if (shiftDown)
+            var isShapeTool = tool is ShapeTool;
+            if (shiftDown && !isShapeTool)
             {
                 bool mouseInLine = MouseCordsNotInLine(mouseMoveCords, thickness);
 
--- a/PixiEditor/Models/Tools/BitmapOperationTool.cs
+++ b/PixiEditor/Models/Tools/BitmapOperationTool.cs
@@ -15,7 +15,7 @@ namespace PixiEditor.Models.Tools
         public bool ClearPreviewLayerOnEachIteration { get; set; } = true;
 
         public bool UseDefaultUndoMethod { get; set; } = true;
-        public virtual bool UsesShift { get => true; }
+        public virtual bool UsesShift => true;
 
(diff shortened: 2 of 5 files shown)


PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
```diff
@@ -128,9 +129,10 @@ private void UseTool(List<Coordinates> mouseMoveCords, BitmapOperationTool tool,
             int thickness = sizeSetting != null ? sizeSetting.Value : 1;
 
             bool shiftDown = Keyboard.IsKeyDown(Key.LeftShift);
-
-            if (shiftDown)
+            var penTool = tool is PenTool;
+            if (shiftDown && !penTool)
             {
+                Debug.WriteLine("shift in BitmapOperationsUtility");
```
flabbet, line 135: Remove the `Debug.WriteLine`, I wouldn't do it as "if tool is not Pen", but rather "if shifDown && isShapeTool"

flabbet (changes requested):

PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
rebizu: wups. forgot to remove it. It should be updated now.

flabbet:
Still, change the condition based on my comment above

rebizu:
> Still, change the condition based on my comment above

sorry I did not see that. It should be fixed now

PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
```diff
@@ -128,8 +128,8 @@ private void UseTool(List<Coordinates> mouseMoveCords, BitmapOperationTool tool,
             int thickness = sizeSetting != null ? sizeSetting.Value : 1;
 
             bool shiftDown = Keyboard.IsKeyDown(Key.LeftShift);
-
-            if (shiftDown)
+            var isShapeTool = tool is ShapeTool;
+            if (shiftDown && !shiftDown)
```
Equbuxu, line 132: 🤔

Equbuxu (changes requested):

PixiEditor/Models/Controllers/BitmapOperationsUtility.cs
Equbuxu: (shiftDown && !shiftDown definitely doesn't look right)

rebizu: copy paste error. sorry. I feel so dum right now.

Equbuxu (changes requested):
So, I've looked through the code, and as it turns out flabbet's suggestion to check for `ShapeTool` doesn't fully solve the issue, as both `PenTool` and `LineTool` are derived from `ShapeTool`. Instead, I think a good solution here would be adding a get-only property to `ShapeTool` that should indicate whether shift should be used with the specific `ShapeTool`. All children of `ShapeTool` should assign a value to that property, showing if that child uses shift. Then, the if statement body should only be executed for `ShapeTool`'s that have that property set. (Right now you have "!isShapeTool" which should be the opposite)

Equbuxu (approved):
Looks good

PixiEditor/Models/Tools/BitmapOperationTool.cs
```diff
@@ -15,6 +15,7 @@ public abstract class BitmapOperationTool : Tool
         public bool ClearPreviewLayerOnEachIteration { get; set; } = true;
 
         public bool UseDefaultUndoMethod { get; set; } = true;
+        public virtual bool UsesShift { get => true; }
```
CPKreuz, line 18: You can simply use this

```suggestion
        public virtual bool UsesShift => true;
```

CPKreuz (changes requested):

PixiEditor/Models/Tools/BitmapOperationTool.cs
rebizu: done

flabbet (approved):

CPKreuz (approved):

PR #213, opened 2021-07-26
Made all converters a markup extension

### LINKED ISSUE #206: Make converters MarkupExtension instead of declaring them all the time https://stackoverflow.com/questions/1659285/should-i-declare-converters-in-app-xaml-or-as-a-per-file-resource

CPKreuz:
Closes #206


PixiEditor/Helpers/Converters/MarkupConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public virtual object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override object ProvideValue(IServiceProvider serviceProvider)
+        {
+            return this;
+        }
```
flabbet, line 24: it can be `=> this`

PixiEditor/Helpers/Converters/NotNullToVisibilityConverter.cs
```diff
@@ -6,13 +6,14 @@
...
+    public class NotNullToVisibilityConverter
+        : MarkupConverter
     {
         public bool Inverted { get; set; }
 
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        public override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            bool isNull = value != null;
+            bool isNull = value is not null;
```
flabbet, line 16: how is this more readable than previous one

flabbet (approved):
I suggest changing stuff mentioned above, but I approve

PixiEditor/Helpers/Converters/NotNullToVisibilityConverter.cs
CPKreuz: I once read that this is a safer way of null check since you can change how `!=` but not how `is not` works

flabbet: Fair point

PR #235, opened 2021-10-15
Added pixiIcon.png and added it to MSIX Package (also changed MSIX ve…

CPKreu:
Added a file icon for pixi files


flabbet:
I think no one will finish this also there is a new logo

PR #240, opened 2021-11-04
Extended crash reports

CPKreuz:
Each crash report is a zip now containing a report.txt and the documents serialized to a .pixi file

The report.txt looks like the old crash report in addition to system information like the CPU, GPU and Memory

I tried adding a lot of (mostly untested) redundancy, for example if the zip file couldn't be created it will create just the txt file

- [x] Add System Information to crash reports
- [x] Add opened documents to crash reports
- [x] Fix memory information
- [x] Add popup when PixiEditor crashes informing the user that it crashed and where he can post the crash report

### DIFF
--- a/PixiEditor/Helpers/CrashHelper.cs
+++ b/PixiEditor/Helpers/CrashHelper.cs
@@ -1,25 +1,134 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using PixiEditor.Helpers.Extensions;
+using PixiEditor.Models.DataHolders;
+using PixiEditor.Parser;
+using PixiEditor.ViewModels;
+using System;
 using System.IO;
-using System.Linq;
+using System.IO.Compression;
+using System.Management;
+using System.Runtime.InteropServices;
 using System.Text;
-using System.Threading.Tasks;
 
 namespace PixiEditor.Helpers
 {
     public static class CrashHelper
     {
-        public static void SaveCrashInfo(Exception e)
+        public static void SaveCrashInfo(Exception exception)
         {
-            StringBuilder builder = new System.Text.StringBuilder();
+            StringBuilder builder = new();
             DateTime currentTime = DateTime.Now;
 
             builder
-                .Append($"PixiEditor crashed on {currentTime:yyyy.MM.dd} at {currentTime:HH:mm:ss}\n\n")
-                .Append("-------Crash message-------\n")
+                .AppendLine($"PixiEditor crashed on {currentTime:yyyy.MM.dd} at {currentTime:HH:mm:ss}\n")
+                .AppendLine("-----System Information----")
+                .AppendLine("General:")
+                .AppendLine($"  OS: {Environment.OSVersion.VersionString}")
+                .AppendLine();
+
+            try
+            {
+                GetCPUInformation(builder);
+            }
+            catch (Exception cpuE)
+            {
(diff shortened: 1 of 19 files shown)


CPKreuz:
Memory information does not work yet and I added a test exception which can be triggered by opening the resize dialog (Document > Resize Document...) and should be remove before merging

flabbet:
How would you like to post crash log? Where it would be posted? GitHub issues?

PixiEditor/Helpers/CrashHelper.cs
```diff
@@ -1,25 +1,134 @@
...
 
 namespace PixiEditor.Helpers
 {
     public static class CrashHelper
     {
-        public static void SaveCrashInfo(Exception e)
+        public static void SaveCrashInfo(Exception exception)
         {
-            StringBuilder builder = new System.Text.StringBuilder();
+            StringBuilder builder = new();
```
flabbet, line 18: You got bonus points for using StringBuilder, gj

CPKreuz:
> How would you like to post crash log? Where it would be posted? GitHub issues?

I think Discord and GitHub (and maybe an email address)

flabbet:
The best would be Discord I guess, email address is way too complicated to do, GitHub could be a way to go, but I am concerned about spam in issues, we would need some kind of checker if issue was already reported. I'd go with discord at first

CPKreuz:
I'm not sure if everyone wants to join a discord server/create a GitHub account

flabbet:
I thought of automated process, without any account creation, like webhook, but this might require a server hmm

CPKreuz:
Well, I could use my raspberry for now

PixiEditor/Helpers/RelayCommand.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace PixiEditor.Helpers
 {
+    [Obsolete("Use GalaSoft.MvvmLight.CommandWpf.RelayCommand instead")]
```
flabbet, line 6: Ummm, why did you mark it as obsolete? I told you we should use our RelayCommand

CPKreuz: We actually said on discord that we were gonna make PixiEditors RelayCommand obselote when I made that commit

And I forgot to remove it again after you said we're gonna use PixiEditor because it supports generics too now

CPKreuz:
Note to future self or anyone working on this branch: Please add PixiEditors version to the report

flabbet (changes requested):
One thing that really bothers me is that Startup window is showing when recovering files

PixiEditor/ViewModels/SubViewModels/Main/WindowViewModel.cs
```diff
@@ -1,15 +1,13 @@
 ﻿using AvalonDock.Layout;
-using PixiEditor.Helpers;
+using GalaSoft.MvvmLight.CommandWpf;
```
flabbet, line 2: This is fine for now, but we will get rid of this after merging #306

flabbet (approved):

PixiEditor/ViewModels/SubViewModels/Main/WindowViewModel.cs
CPKreuz: alright

PR #302, opened 2021-12-19
Command Manager

CPKreuz:
The Command Manager is supposed to allow better access to all the commands that are accessible to the user.

It will use Attribute's for defining commands

I'm planning on adding these things in this PR

- [x] Custom Shortcuts ([Seems to be already implemented here, but I wanted to approach it a bit differently](https://github.com/PixiEditor/PixiEditor/pull/228/files))
- [x] Custom Shortcuts Window
- [x] Command Search Window ([Similar to what VS Code has](https://user-images.githubusercontent.com/45312141/146681750-4680af37-839c-4135-95a1-d772c03a08c7.png))
 
Left to fix:

- [x] Shortcuts Window
- [x] Re-implement holding tool shortcut to use tool temporarily

Implemented Commands

- [x] Basic Command (Simple command Parameter)
- [x] Tool Command (Allows to add a tool command, will use the SelectToolCommand from ToolsViewModel)


Equbuxu:
- Crash on launch if shortcuts.json doesn't exist
- The command manager gets stretched to fit it's contents. If one the the recent files has a really long filename the command manager will get stretched all the way to the edges of the screen
- Would be nice if you could close the window on Escape
- For some reason there is a crash when trying to load a recent file that doesn't exist anymore, you don't have to use the command manager for that.
- You shouldn't be able to bind multiple actions to the same key
- Alt should be bindable on it's own (not as a part of a key combination)
- A reset all shortcuts button might come in handy
- Ctrl+Shift+Alt gets displayed as Alt+Ctrl+Shift, alt being in the beginning feels kinda weird
- Why is backspace called "Back"?

CPKreuz:
note to future self, add command for toggeling pen mode

CPKreuz:
also don't forget to implement debug mode and more copy color commands future self

PixiEditor/Models/Commands/CommandController.cs
```diff
@@ -0,0 +1,322 @@
...
+                        "shortcuts.json"),
+                    this);
+
+            Commands = new();
+            CommandGroups = new();
+            CanExecuteEvaluators = new();
+            IconEvaluators = new();
+        }
+
+        public void Init(IServiceProvider services)
```
Equbuxu, line 45: very scary looking function, it could use some splitting up

Equbuxu (changes requested):
- [x] You can't choose items in the command manager by clicking them, the window gets closed when you do
- [x] There is a crash when trying to load a latest file that doesn't exist anymore. I don't know how this PR is causing it, but it's not present in the commit that is right before the first one in this PR.
- [x] There is a typo in "asigned" in the window that pops up when you try to use the shortcut that already exists. Also I don't understand what Replace and Switch mean, they both seem to do the same thing?
- [x] The debug commands (Crash application, delete preferences, etc.) are visible in the command window in release build
- [x] When you first open the keybinds tab in settings there is a "nothing found" text at the very top
- [x] Let's say you are typing in the command window and have your mouse over it by accident. Then you press the down arrow to select the option you want. The selection moves down from where your mouse is, but you wanted to select the first item in the list.
- [x] I'd prefer the command manager to always have a fixed size. Even though it doesn't stretch all the way to the edges of the screen anymore it's width still jumps around as you type and see different options.

https://user-images.githubusercontent.com/29402050/170100031-701a2a06-9f6d-41e2-82b3-00275c1ca075.mp4

- [ ] You can't still bind Alt by itself, but I guess it's fine since the transient shortcuts aren't bindable in general
- [ ] The rgb and rgba commands are impossible to find unless you know about them, since their item only gets displayed after you've fully typed the command
- [x] Reset shortcuts (not really required) and set your undo shortcut to "C". In the window, choose "Replace". If you now press C, it will undo as expected (I think? What does "Replace" mean again?). Now restart the app. The C key now selects the circle tool and calls undo at the same time.
- [x] The shortcuts window should have a hint saying that you can change them in settings
- [x] Very valid color
very valid color

flabbet:
I miss 2 things here. One is the search bar on top (Something like VS Community has), which brings search window, it took me a while to find out that `CTRL+K` opens it.

Second is keybinding templates, like user can pick ready-made template, but this can be a task for another day.

Generally it's really cool, I like it. Both shortcuts and search window. I'll do proper code review in sunday probably.

Equbuxu:
Note to myself or someone else: check what happens when you press "open crash reports directory" when the crash_logs directory doesn't exist

CPKreuz:
> Note to myself or someone else: check what happens when you press "open crash reports directory" when the crash_logs directory doesn't exist

@Equbuxu 

It crashes, but it's self fixing since the location will be created after the crash

So no need to fix it if it's self repairing
ez

flabbet (approved):

Equbuxu (approved):

PR #303, opened 2021-12-21
Rect based undo

flabbet:
yeah baby. Let's go. Also I left dispose there @Equbuxu look at it please

### DIFF
--- a/PixiEditor/Models/Undo/StorageBasedChange.cs
+++ b/PixiEditor/Models/Undo/StorageBasedChange.cs
@@ -342,11 +342,17 @@ namespace PixiEditor.Models.Undo
             int targetWidth = widthBigger ? chunk.Width : layer.Width;
             int targetHeight = heightBigger ? chunk.Height : layer.Height;
 
+            targetWidth = Math.Clamp(targetWidth, 0, layerData.MaxWidth);
+            targetHeight = Math.Clamp(targetHeight, 0, layerData.MaxHeight);
+
             int offsetDiffX = layerData.OffsetX - layer.OffsetX;
             int offsetDiffY = layerData.OffsetY - layer.OffsetY;
 
             int targetOffsetX = widthBigger ? layerData.SerializedRect.Left : layerData.OffsetX;
             int targetOffsetY = heightBigger ? layerData.SerializedRect.Top : layerData.OffsetY;
+
+            targetOffsetX = Math.Max(0, targetOffsetX);
+            targetOffsetY = Math.Max(0, targetOffsetY);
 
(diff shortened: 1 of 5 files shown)


PixiEditor/Models/Undo/StorageBasedChange.cs
```diff
@@ -14,45 +16,64 @@ namespace PixiEditor.Models.Undo
     /// </summary>
     public class StorageBasedChange : IDisposable
     {
-        public static string DefaultUndoChangeLocation => Path.Join(Path.GetTempPath(), "PixiEditor", "UndoStack");
+        public static string DefaultUndoChangeLocation { get; } = Path.Join(Path.GetTempPath(), "PixiEditor", Guid.NewGuid().ToString(), "UndoStack");
```
CPKreuz, line 19: Maybe there should be a subfolder for the sessions, something like `Temp/PixiEditor/Sessions/GUID/UndoStack`

CPKreuz (commented):
When is the undo stack deleted?

PixiEditor/Models/Undo/StorageBasedChange.cs
Equbuxu: Actually the only purpose of the PixiEditor folder is for us to have an easy way to find it during debugging

So I don't see why we would need another folder

Equbuxu:
> When is the undo stack deleted?

The individual images are deleted in storagebasedchange's dispose which is called whenever the change is no longer used (including on document/app) close. If for any reason the images don't get deleted they just stay there until windows clears the temp folder

PixiEditor/Models/Undo/StorageBasedChange.cs
CPKreuz: Well, it would make it easier to debug, also I think that the session folder should be deleted when PixiEditor closes

Equbuxu (changes requested):
Broken

https://user-images.githubusercontent.com/29402050/148635687-6c06a3b1-e47f-466c-aa1b-af9dfb7fd959.mp4

flabbet:
Fixed

Equbuxu (changes requested):
Broken

https://user-images.githubusercontent.com/29402050/148654943-5dbf5e2b-751b-4db6-bccf-8fccf078465e.mp4

flabbet:
Try me @Equbuxu one more time

Equbuxu:
Well

https://user-images.githubusercontent.com/29402050/149121440-f50d2cd1-6fe7-4432-9f74-74db06d85290.mp4

Equbuxu (approved):

PR #306, opened 2021-12-30
Palettes

flabbet:
Left TODO: 

[] - Lospec palette browser
[] - Update palette color
[] - Update changed color in document


flabbet:
Added compact mode: 
image

flabbet:
Finished! New version of PixiParser must be uploaded first, then this can be reviewed properly

README.md
```diff
@@ -82,7 +82,7 @@ Struggling with something? You can find support in a few places:
 
 ### Software Requirements
 
-* .NET 6
+* .NET 5
```
Equbuxu, line 85: huh?

Equbuxu (changes requested):
- The icons are pretty confusing, so I propose some changes:
  - Palette window:
- [x] "Browse palettes" (currently a folder icon) -> [magnifying glass](https://images.google.com/search?q=browse+icon)
- [x] "Load palette" (currently an upload icon (why??)) -> [folder icon](https://images.google.com/search?q=open+file+icon)
- [x] "Replace color" (currently a shuffle icon) -> [replace icon](https://www.flaticon.com/premium-icon/replace_754616?term=replace&page=1&position=1&page=1&position=1&related_id=754616&origin=search)
  - Palette browser
- [x] "Import" (currently a download icon) -> checkmark icon (not sure about this one, but the current download one feels like "save this palette as a file")
- The palette browser is (imo) missing an import from file option (please don't use the download icon, the import one has a fully enclosed box, something like [this](https://www.flaticon.com/free-icon/import_243718?term=import&page=1&position=22&page=1&position=22&related_id=243718&origin=search))
- When the palette browser is empty, it would be nice to display something like "you can download palettes from lospec or save the current palette here or ... etc".
- The entire window has the old dialog style (empty title bar, big everything). Would be nice to scale it down a bit, or at the very least put some text into the title bar
- [x] In addition to the previous point, the checkboxes are kinda giant in the whole app now (especially noticeable in settings) 
- [x] Slight inconsistency, the delete palette button gets highlighted on hover, but not any of the other ones. It's not really an issue, but I really like the highlighting, maybe we should add it to all the buttons?
~- The colors are way too easy to drag. If you shift your mouse by even a single pixel while clicking the click won't register and you won't choose the color.~ Idk how to fix this one, I just added "Choose" option in context menu
- [x] Why is the color to be replaced a square, but the color it's going to be replaced with is a long rectangle?
- [x] Proposal: Let's make the swatches and the palette colors smaller (and reduce margins/padding?). I just wanna see all of them at the same time, and right now they are more than big enough to hit comfortably
~- Proposal: Make the top panel where all the buttons are collapsible (same as the bottom one). Wanna see more colors.~
- Okay, I've just realized that I can dock and compress the palette window to see more colors, but only after reading the code and remembering about the screenshots you've sent in discord. No one will figure out you can do that on their own
- There is no failure message when pixieditor fails to download a palette from lospec

----

- [x] Crash when renaming a palette if a palette with this name already exists
- [x] If a palette downloaded from lospec has the same name as one of your existing ones, it gets overridden without warning
- [x] Rename one of your palettes to "NAIVE39" (all caps). Get [this](https://lospec.com/palette-list/naive39) palette from lospec. Your palette gets replaced with the downloaded one, but the downloaded one is also added to the list. So now there are two entries for the same palette.
- [x] Heart.png is unused (I think)
- [x] Open the directory where the palletes are stored via the convenient button and delete one of  the files. The window doesn't notice anything, not even if you reopen it. Try renaming the deleted palette in the window to crash the app.
- [x] Crash when giving a palette an extremely long name
- [x] You can rename your palette to a bunch of spaces (or give it an empty name) and it will appear to have worked, but the name won't be saved.
- [x] By default, the color to be replaced is transparent. This lets you replace all the transparent pixels with some color. It works as most of the time (it doesn't affect the area outside the dynamic bitmap), however, if you try to undo it you'll have to press undo twice. Even more interestingly, if you had say a black color in the palette, replaced everything with black, and then pressed undo, the black color in the palette is replaced by a transparent color.

CPKreuz:
> "Load palette" (currently an upload icon (why??)) -> [folder icon](https://images.google.com/search?q=open+file+icon)
>   - Palette browser
> "Import" (currently a download icon) -> checkmark icon (not sure about this one, but the current download one feels like "save this palette as a file")


I still think that the download icons should be replaced with a plus icon

flabbet:
I improved almost everything from requested changes. I commented @Equbuxu comment with changes. I didn't do anything to automatically bring paletteviewer into side with small stuff. I have no clue how to do that with AvalonDock, so let's leave that for later.

Equbuxu (changes requested):
- Please fix warnings
- "The colors are way too easy to drag. If you shift your mouse by even a single pixel while clicking the click won't register and you won't choose the color" is actually a big issue. When using a drawing tablet I have to click like 4-5 times before I can finally pick up the color, because it's hard to keep the pen completely stationary.
- It's a bit weird how only one of the rectangles has rounded corners
image
- The "open folder containing palettes" button is missing a mouse hove hint thingy, same for the "add to favourites" (star) button and the "edit palette name" (pen) button
- Collapsing the replace color menu leaves behind a gray background covering the palette colors
image
- All interaction with the palette window browser becomes slow once you have a lot of palettes (about 20). Seems really similar to what happens to layers.
- Mark a palette as favorite and then delete it using explorer. Press the "add from current palette" button (you need to have some colors so that there is stuff to add). The app crashes.
- This one is ridiculous, but every time you have a name collision, one of the palettes gets "(1)" appended to it's name. If a palette with that name exists too, another "(1)" is appended, meaning you can infinitely stack them. So, if you name 72 pallettes "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" the stack of "(1)" will reach the filename size limit and pixieditor will crash. I don't know why anyone would do that. I don't know why I did it. The palette browser became so slow I had to wait like 5 seconds after each rename

flabbet:
I'll ignored some of the visual stuff for now, same with performance. I'll try to improve it later. Right now I'd prefer to focus on other PixiEditor stuff

Equbuxu (commented):
A pretty obscure crash, but if you open palette browser, open folder containing palettes, select one of the palette files, copy it, and hold down ctrl+v for a few seconds pixieditor might crash saying that the file is currently being used by another process. If it doesn't crashes you can also select all the palettes you've created and delete them. Now pixieditor might crash with a message saying file not found

flabbet:
Fixed! I left some warnings because I found them useless

Equbuxu (changes requested):
- [x] There is a binding error
- [x] Importing the same palette from lospec multiple times bring up the "override palette?" dialog, but when you click yes the palette gets duplicated. Reopening the window removes the duplicate.
- [ ] I got a random crash in
```csharp
private void HandleCachePaletteCreated(Palette updatedItem)
{
    SortedResults.Add(updatedItem); // here
    Sort();
}
```
I wasn't smart enough to save the actual error message, and I can't replicate it now, so idk if you will be able to fix it. We can ignore it for now and wait for it to happen again.

flabbet:
Seems about right. Let's merge and publish dev release for users to test

Equbuxu (approved):

Issue #295, opened 2021-12-17
Move Viewport tool should be always be replaced with previous tool on space key release


Equbuxu:
Pressing and releasing space should always bring up the previous tool, even if you haven't moved the viewport. Middle mouse button works this way already. The normal shortcut that lets you switch permanently can be different

Issue #130, opened 2020-12-29
Add percentage/fraction input when changing document/canvas size


Equbuxu:
Sometimes when changing image/document size you want to make it twice as big, or twice as small, or something similar. It would be useful to have an option to do this easily. One way to implement this would be making a dropdown with different units (pixels, percents/fraction, other can be added as well later)

tomaszkot:
What do you think about a Radio Box approach, like : 

![image](https://user-images.githubusercontent.com/2432291/149661444-96c3efd4-f7f8-45e1-84df-76d30e7a0b35.png)

flabbet:
Yeah, this has some potential. We would need to style it quite nicely

Equbuxu:
Percentage should be separate for width and height just like pixels. I would add a dropdown titled "units" where you can choose pixels or percents

tomaszkot:
> Percentage should be separate for width and height just like pixels.

If we really go into that direction, then yes combo is a way to go. I assumed most of the time user choosing 'Percentage' wants them the same in both axes.  In the long run combo approach is more flexible.

Equbuxu:
Now that I think about it a single percentage field might be fine too, having different percentages is a rare use case. But then looking at your example the radio box is kinda redundant, as the values in the other field can be recalculated when the percentage changes (similar to the way those fields behave when Preserve Aspect Ratio is enabled).

@tomaszkot do you have any plans to work on this issue or should I take a crack at it

tomaszkot:
As @flabbet mentioned the other option would be disabled. If you do not provide radio boxes user would be confused which option would be applied.

I can work on that issue at the end of the week. If you can do it earlier or just feel like doing it - fell free to do so.

tomaszkot:
I finally did it :)

Equbuxu:
> I finally did it :)
I've done quite a bit of work updating dialogs (but haven't worked on this issue yet). Could you try to rebase your branch onto dialogs-revamp and PR into there? Hopefully you won't have too many conflicts, but if you do then PR into master and I will take care of that myself

Equbuxu:
> I finally did it :)

Actually, I didn't realize you've used the same branch as in your previous PR (that's kinda messy). Disregard what I said about rebasing

tomaszkot:
[...]

PR #312, opened 2022-01-16
#295 Restore a tool when Space is released

tomaszkot:
#295 Restore a tool when Space is released

The change I made helps as long as you do not change an active tool. Once you do it  IoViewModel.ProcessShortcutDown is no longer called. Clearly some focus-related issue. Interestingly enough if you focus a layer at right side that problem with focus is not happening. So it looks like  a bigger issue of DrawingViewPort being defocused depend on user action. Seems like refocusing it back at the end of tool changing event handler would do the job but I do not know WPF well enough to do it.

### DIFF
--- a/PixiEditor/ViewModels/SubViewModels/Main/ToolsViewModel.cs
+++ b/PixiEditor/ViewModels/SubViewModels/Main/ToolsViewModel.cs
@@ -176,7 +176,8 @@ namespace PixiEditor.ViewModels.SubViewModels.Main
         private void SetActiveTool(Type toolType)
         {
             Tool foundTool = ToolSet.First(x => x.GetType() == toolType);
-            SetActiveTool(foundTool);
+            if(ActiveTool != foundTool)//otherwise we would lost track of LastActionTool
+              SetActiveTool(foundTool);
         }
 
         private void SetToolCursor(Type tool)


flabbet:
I don't think you actually solved the issue. Press space once, it will not bring previous tool as active.

tomaszkot:
I just fixed the edge case of the first usage. Yet, my original comment about loosing a focus is still valid.

tomaszkot:
Not sure why it failed to build, but is the fix resolving the issue (assuming focus in set) ?

Equbuxu:
#295 still isn't fixed properly. With your implementation, all of the tool shortcuts stop working, e.g. if you press B to switch to the pen tool it switches right back on key release. To clarify the issue, using space to bring up the move viewport tool is a special behavior. It is different from how the shortcuts normally work, and at the moment the shortcut system does not support it. Because of that, MoveViewportTool's shortcut should be changed to a letter (H). H would work like every other shortcut. The space key behavior is then implemented separately inside the IoViewModel. Using the space key is basically the same as pressing the middle mouse button, which is implemented inside IoViewModel's OnMouseUp and OnPreviewMiddleMouseButton.

#130 needs more work. Document -> Resize Document dialog has "1" as the default percentage. Typing in a large value like 99999 and resizing causes a crash. The UI looks awkward, the textbox isn't centered relative to the "percentage" label and the label itself is offset from the radio button. The radio buttons need to be styled to fit the rest of the program.

ps: please make separate PRs for different issues in the future

tomaszkot:
Sorry about putting two tasks inside one PR. I reverted commits for #295, I have to redo it this PR will be used for #130. 

> Document -> Resize Document dialog has "1" as the default percentage

I did it in my last commit, other issues will be fixed soon.

flabbet:
If you change percentage, for example to 50%, unfocus and then change to 100% it doesn't come back to original width/height

Equbuxu:
We merged a different branch into master and now there are a bunch of conflicts, I can help you with them if they are too difficult to resolve

tomaszkot:
> If you change percentage, for example to 50%, unfocus and then change to 100% it doesn't come back to original width/height

Thx for info, I fixed it.

tomaszkot:
> We merged a different branch into master and now there are a bunch of conflicts, I can help you with them if they are too difficult to resolve

I'm not a git champion, using other  VCS on daily basis. 
Would it be smth like:

image

Equbuxu:
> > We merged a different branch into master and now there are a bunch of conflicts, I can help you with them if they are too difficult to resolve
> 
> I'm not a git champion, using other VCS on daily basis. Would it be smth like:
> 
> image

Yeah, that's one of the ways to do it (don't forget to pull before merging).

tomaszkot:
Uff, that was hard to merge...

flabbet:
Why can't user go higher than 2048px in resize in precents?

flabbet:
Maximum size is 9999x9999

tomaszkot:
> Why can't user go higher than 2048px in resize in precents?

I initially planned to use 9999, but since it's not a power of two I started looking around in the code and found out RecentlyOpenedDocument has limitation of 2048, so decided to use it. ImageFileMaxSizeChecker also assumes it. I think 2048 is a resolution that should work reasonably fast on average comp. I think we could try out 4096.

PR #314, opened 2022-01-21
Fixes

flabbet:
- Updated PixiParser and AvalonDock
- Fixed how stride is caluclated, when converting pasted bitmaps
- Fixed paste infinite undo/redo adding loop
- Changed target format of unsupported PNG color formats to Bgra32 instead of Rgba64 (Skia had problem with that for some reason)
- Removed focus outline for grids


Equbuxu:
A crash I found on this branch:
- Draw something
- Select it
- Ctrl+C
- Remove selection
- Delete base layer
- Draw something
- Paste
- Undo x3
- Redo
- Draw something

Not sure if you want to fix it here, you can move this into an issue

Equbuxu (approved):
Looks good to me so this can be merged if you don't want to fix anything else here

Issue #278, opened 2021-12-10
Support more export formats


Equbuxu:
Currently, PixiEditor only supports .png and .pixi file formats. It would be nice to add more ones, like .jpg, .bmp, .tiff and .webp

tomaszkot:
I can do is as a common code along with #277

tomaszkot:
I did a first attempt to it, but when pressing 'Creating Pull Request' in VS it still appends to existing PR instead of creating a new one :/.

Equbuxu:
> I did a first attempt to it, but when pressing 'Creating Pull Request' in VS it still appends to existing PR instead of creating a new one :/.

A pull request is just a request to merge a branch. Github just shows all of the commits you wanna merge. So if you commit to the same branch those commits will show up in the same PR.

Now, I see that you've managed to make a separate PR which is #319, but there is a problem with it because it contain all commits from #312. At the momen

Equbuxu:
All of this makes me think we should improve our contribution manual because there is a simple rule that can help avoid all of your issues: before working on a feature make a new branch for it based on master. Never commit directly to master. You've already commited to master in #312 and because of that you can't really make branches based on your master anymore as they will contain those commits

Equbuxu:
By the way, github isn't a very convenient place to discuss this, if you use discord you should join our chat https://discord.gg/qSRMYmq, that's where most of the development discussions happen

flabbet:
Yup, we will merge #319 when #312 will be ready to merge

PR #319, opened 2022-01-27
#278 Support more export formats 

tomaszkot:
1st  attempt  to exporting of many formats.


flabbet (changes requested):
It does not compile :p

flabbet (changes requested):
Yeah, resizing is good. But now please take into consideration, that you can export images into different formats, but still can only open .png, .jpg and .pixi. If we want to support more formats, PixiEditor needs to handle exporting and importing as well.

tomaszkot:
> Yeah, resizing is good. But now please take into consideration, that you can export images into different formats, but still can only open .png, .jpg and .pixi. If we want to support more formats, PixiEditor needs to handle exporting and importing as well.

That code was supposed to implement #278. Task was about exporting. I'd rather do import as a separate issue/PR but it you say we shall continue in that one I can give it a try.

flabbet:
You know what, yeah. You're right. Let's do it in another PR

PixiEditor/Models/IO/Exporter.cs
```diff
@@ -48,22 +52,35 @@ public static string SaveAsEditableFile(Document document, string path)
             return path;
         }
 
+        //static Dictionary<ImageFormat, Action<ExportFileDialog, WriteableBitmap>> encoders = new Dictionary<ImageFormat, Action<ExportFileDialog, WriteableBitmap>>();
+        //TODO remove static methods/members
+        static Dictionary<ImageFormat, Func<BitmapEncoder>> encodersFactory = new Dictionary<ImageFormat, Func<BitmapEncoder>>();
```
flabbet, line 57: This is weird, can't you just create Dictionary of <ImageFormat, BitmapEncoder> and instantiate them if null?

flabbet (changes requested):

PixiEditor/Models/IO/Exporter.cs
tomaszkot: Both approach would do the job. Current one creates a transient BitmapEncoder, in your proposal it will stay in memory. 
Since memory footprint is small here I would change to your approach as it's a bit simpler code there.

tomaszkot:
I added  more options Save/SaveAs dialog. Now what remains is open dialog with them. I can not do it on a new feature branch until you merge this one, as I would share some code with it.

tomaszkot:
On the other hands I could try to do as part of this RP though I'm not sure how much changes it involves...

flabbet (approved):

Issue #332, opened 2022-02-03
crash - when recent file not exists


tomaszkot:
When recent file not exists

![image](https://user-images.githubusercontent.com/2432291/152386709-c7d57749-80a5-4108-97fd-21d0f2aeb66a.png)

Issue #316, opened 2022-01-23
Recently used files - can I remove items from it?


tomaszkot:
I deleted files I recently used but the keep popping up on HelloTherePopup dialog and are available in File->Recent menu. When clicking one in the menu it disappears from the menu but after restarting app it's again there.

Equbuxu:
It's strange that they come back after disappearing. I think right now there is no way to clear the list or to remove individual items, it would be nice to add these features at some point.

CPKreu:
This was done with 6d1bbc7ec73f6506c8adf024f45a16d57129489b

PR #333, opened 2022-02-03
crash - when recent file not exists #332

tomaszkot:


CPKreuz (commented):
Looks good

If you want to contribute a bit more you could remove the logic that removes recently opened document's when they are not found like mentioned in #316

It should not be removed in my opinion as it is possible for files to become unavailable for a temporary time (e.g. a network drive could not be mapped)

flabbet:
Yep, this is nice. If you want, you can do the thing @CPKreuz mentioned

flabbet (approved):

PR #335, opened 2022-02-04
The selection does not prevent you from drawing outside of it

tomaszkot:
Functionality done, no UT so far...

### DIFF
--- a/PixiEditor/Models/Controllers/BitmapManager.cs
+++ b/PixiEditor/Models/Controllers/BitmapManager.cs
@@ -10,6 +10,7 @@ using SkiaSharp;
 using System;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
+using System.Linq;
 using System.Windows;
 
 namespace PixiEditor.Models.Controllers
@@ -176,6 +177,9 @@ namespace PixiEditor.Models.Controllers
         {
             if (activeSession == null)
                 throw new Exception("Can't execute tool's Use outside a session");
+
+            if (!activeSession.MouseMovement.Any())
+                return;
 
             if (activeSession.Tool is BitmapOperationTool operationTool)
             {
--- a/PixiEditor/Models/Controllers/BitmapManager.cs
+++ b/PixiEditor/Models/Controllers/BitmapManager.cs
@@ -10,7 +10,6 @@ using SkiaSharp;
 using System;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
-using System.Linq;
 using System.Windows;
 
 namespace PixiEditor.Models.Controllers
@@ -177,9 +176,6 @@ namespace PixiEditor.Models.Controllers
         {
             if (activeSession == null)
                 throw new Exception("Can't execute tool's Use outside a session");
-
-            if (!activeSession.MouseMovement.Any())
-                return;
 
             if (activeSession.Tool is BitmapOperationTool operationTool)
             {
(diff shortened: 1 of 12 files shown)


Equbuxu:
I haven't tested the branch, but looking through the code I don't think your approach will work with a brush size greater than 1. The shape tools probably won't function properly either

tomaszkot:
image

I draw the line across the whole image. There are some pixels out of selection due to the tool size. Do you meant that problem ?

Equbuxu:
> image
> 
> I draw the line across the whole image. There are some pixels out of selection due to the tool size. Do you meant that problem ?

Yes

tomaszkot:
> > image
> > I draw the line across the whole image. There are some pixels out of selection due to the tool size. Do you meant that problem ?
> 
> Yes

OK. I think the simplest approach would be to paint on a temporary and then paste result cropped to the Selection.
I did following  try (tmpLayer  is a member):

 public override void Use(Layer activeLayer, Layer previewLayer, IEnumerable<Layer> allLayers, IReadOnlyList<Coordinates> recordedMouseMovement, SKColor color)
        {
         ...
            if (tmpLayer == null)
              tmpLayer = new Layer("tmp", previewLayer.Width, previewLayer.Height);
            CreateRectangle(tmpLayer, color, fillColor, recordedMouseMovement, thickness);
            tmpLayer.LayerBitmap.SkiaSurface.Draw(previewLayer.LayerBitmap.SkiaSurface.Canvas, 0, 0, Surface.BlendingPaint);
        }
        
        But noting is drown...

Equbuxu:
> > > image
> > > I draw the line across the whole image. There are some pixels out of selection due to the tool size. Do you meant that problem ?
> > 
> > 
> > Yes
> 
> OK. I think the simplest approach would be to paint on a temporary layer and then paste result cropped to the Selection. I did following try (tmpLayer is a member):
> 
> public override void Use(Layer activeLayer, Layer previewLayer, IEnumerable allLayers, IReadOnlyList recordedMouseMovement, SKColor color) { ... if (tmpLayer == null) tmpLayer = new Layer("tmp", previewLayer.Width, previewLayer.Height); CreateRectangle(tmpLayer, color, fillColor, recordedMouseMovement, thickness); tmpLayer.LayerBitmap.SkiaSurface.Draw(previewLayer.LayerBitmap.SkiaSurface.Canvas, 0, 0, Surface.BlendingPaint); }
> 
> ```
>     But noting is drown... 
> ```

The concept of a temporary layer actually already exists, it's the preview layer. It's drawn on the target layer once you are done using the tool. (I think it's called ApplyPreviewLayer, use ctrl+t to find it) I think it would make the most sense to clip to selecion there. It would create visual artifacts, but changing every single tool to account for selection would be too much.

We also already have a function that clips an image to selection, I don't remember it's name but it is used when copying to clipboard.

And one last thing, please avoid using the selected pixels array if you can (it might be a good idea to mark it as deprecated) as we are planning to get rid of it. Use the selection layer image wherever possible

tomaszkot:
I saw ApplyPreviewLayer, but my understanding was we do not want the red part from beneath be rendered:

image

That was not possible using ApplyPreviewLayer. I did it with a (hopefully) minimal effort.

tomaszkot:
Hmm, I just realized there is a circle selection to be also considered...

flabbet:
Don't waste time doing it this way. We are dropping support for selected points soon. This has to be done using preview selection layer and masking

flabbet:
In fact, besides circle selection, you can make _any_ selection shape by using `Add` mode

PR #336, opened 2022-02-07
Move Viewport tool should be always be replaced with previous tool on space key release #295

tomaszkot:
MoveViewportTool assigned to H key.
MoveViewportTool transient on assigned to Space key.

### DIFF
--- a/PixiEditor/Models/Tools/Tools/MoveViewportTool.cs
+++ b/PixiEditor/Models/Tools/Tools/MoveViewportTool.cs
@@ -1,5 +1,4 @@
 ﻿using PixiEditor.Models.Position;
-using PixiEditor.ViewModels.SubViewModels.Main;
 using System.Collections.Generic;
 using System.Windows.Input;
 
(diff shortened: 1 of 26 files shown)


flabbet (approved):

Equbuxu (changes requested):
Good job, this is going in the right direction. One issue: 
1. Select Move Viewport tool manually (using mouse)
2. Move the Viewport a bit
3. Press and release middle mouse button or space

This brings up the pencil tool which was selected in the very beginning. Since the Move Viewport tool was already selected before step 3, pressing space and middle mouse button should have no effect

tomaszkot:
> Good job, this is going in the right direction. One issue:
> 
> 1. Select Move Viewport tool manually (using mouse)
> 2. Move the Viewport a bit
> 3. Press and release middle mouse button or space
> 
> This brings up the pencil tool which was selected in the very beginning. Since the Move Viewport tool was already selected before step 3, pressing space and middle mouse button should have no effect

Yes, I just fixed it.

tomaszkot:
> Looks good to me, there is a small detail I forgot about last time. There is a tooltip that has a small description of the tool when you hover your mouse over it. That description still says that the shortcut is Space instead of H. It needs to be updated manually (which kinda sucks)

Yes, I made them being read dynamically. So now there is nothing blocking such enh: #343

PixiEditor/Models/Tools/Tools/MoveViewportTool.cs
```diff
@@ -18,7 +18,13 @@ public MoveViewportTool(ToolsViewModel toolsViewModel)
         }
 
         public override bool HideHighlight => true;
-        public override string Tooltip => "Move viewport. (Space)";
+        public override string Tooltip
+        {
```
flabbet, line 22: Can you make it the same way as other tools? Arrow function instead of whole `get` body?

flabbet (approved):

Equbuxu (approved):

Issue #337, opened 2022-02-09
Preserve aspect ratio doesn't work like it supposed to 


flabbet:
1. Open new file dialog
2. Set width to 32
3. Select "Preserve aspect ratio"
4. change width to 64

Expected: Width: 64, Height: 128
Actual: Width:64, Height: 64

CPKreuz:
It does work in the resize canvas dialog

Equbuxu:
fun fact: paint.net behaves exactly the same as pixieditor does now. Other apps don't have the maintain aspect ratio option at all (at least when making a new document)

flabbet:
Related files: 

https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/Views/Dialogs/NewFilePopup.axaml.cs
https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/Views/Dialogs/NewFileDialog.cs
https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/Views/Input/SizePicker.axaml.cs

PR #379, opened 2022-02-28
Fixed restoring documents

### LINKED ISSUE #357: PixiEditor fails to load recovered files even though they were successfully saved
1. Open this file [before_clip.zip](https://github.com/PixiEditor/PixiEditor/files/8098952/before_clip.zip)
2. Draw something
3. Crash
4. Press recover files
5. PixiEditor opens with no files, even though the file was saved successfully and loads fine if you extract and open it manually

flabbet:
Fixes #357

### DIFF
--- a/PixiEditor/Models/DataHolders/CrashReport.cs
+++ b/PixiEditor/Models/DataHolders/CrashReport.cs
@@ -91,8 +91,9 @@ namespace PixiEditor.Models.DataHolders
 
         public int GetDocumentCount() => ZipFile.Entries.Where(x => x.FullName.EndsWith(".pixi")).Count();
 
-        public IEnumerable<Document> RecoverDocuments()
+        public List<Document> RecoverDocuments()
         {
+            List<Document> documents = new();
             foreach (ZipArchiveEntry entry in ZipFile.Entries.Where(x => x.FullName.EndsWith(".pixi")))
             {
                 using Stream stream = entry.Open();
@@ -109,8 +110,10 @@ namespace PixiEditor.Models.DataHolders
                     continue;
                 }
 
-                yield return document;
+                documents.Add(document);
             }
+
+            return documents;
         }
 
         public void Dispose()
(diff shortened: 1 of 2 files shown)


CPKreuz:
Didn't fix #357 for me, maybe we should add an popup displaying errors when recovering documents
And maybe wording `might be recoverable` in crash dialog would be better

Equbuxu (changes requested):
- The collapse button of the reference layer is still focusable, the same goes for:
  - The swap colors button inside of the color picker, not sure if this can be fixed without 
  changing the color picker.
  - HEX/RGBA and background buttons of the navigator window
  - The close layer button
  - The pen mode button
- Layer opacity undo breaks after using the scroll wheel (actually maybe in general too)
  1. Make a new document
  1. Scroll down above the opacity field
  1. Hold down undo
  1. It cycles between two values instead of going back up to 100
- Pasting an image does strange things when the image is bigger than the canvas
  1. Copy a 64x64 image
  1. Paste it on a 16x16 canvas
  1. The image gets cropped and now you've got a big highlight
- Exception when loading recovered files even though the files aren't corrupted
  1. Download [crash-2022-03-04_02-39-31_420.zip](https://github.com/PixiEditor/PixiEditor/files/8181939/crash-2022-03-04_02-39-31_420.zip)
  1. PixiEditor Project Properties -> Debug -> General -> Open debug launch profiles UI -> Command line arguments -> Set to `-- crash "pathtofile"`
  1. Launch and recover files to see an exception (ViewModelMain.Current is null)
- #357 still isn't fixed, but I fixed it in https://github.com/PixiEditor/PixiParser/pull/14. Once it is merged we'll need to update the package

Equbuxu (changes requested):
- The close \*tab* button (I've called it the "close layer button" by mistake last time) is still focusable.
- I'd argue that the fix to #337 also broke the preserve aspect ratio function in the resize image/canvas dialogs, cause there I would expect them to always use the canvas proportions, not the stuff you've typed

Equbuxu (approved):
I found another case where undoing pasting doesn't bring selection back but I decided to make it into an issue (#394). I've also made an issue for recovering files after crash (#393). I think we should merge this PR and make a dev release

Issue #140, opened 2021-01-13
Windows explorer image preview


Equbuxu:
Show a preview of .pixi files in windows explorer, similar to regular images.

tomaszkot:
I think I could do it using SharpShell (https://github.com/dwmkerr/sharpshell). It has MIT licence so I suppose can be used in PixiEditor.

CPKreuz:
I think I tried it using that before
I think it only worked with .NET Framework and I didn't get it to work

But have fun

flabbet:
Yeah, I would love to see it in action

tomaszkot:
> I think I tried it using that before I think it only worked with .NET Framework and I didn't get it to work
> 
> But have fun

Indeed it's not working when compiled to .net6 :/

tomaszkot:
Almost there
![image](https://user-images.githubusercontent.com/2432291/156879946-4d689237-27f9-437a-b03a-ae3d243cb9d7.png)

PR #389, opened 2022-03-05
Windows explorer image preview #140

tomaszkot:
This is the 1st working approach.
This is a native code relaying on the PixiEditor making a temporary image during save. See changes in Exporter SaveAsEditableFile. 
Since PixiEditor is an application I can not reuse its code to load a thumbnail preview.
Maybe in the future app's code loading images could be extracted to a dll, then it would be possible to use that code from a native code.
So, the current limitation is that if someone makes a copy of the pixi file in windows thumbnail  will not be changes until file is opened and saved in the application.

To test it: Go to \Installer\PixiThumbnails and run InstallPixiThumbnailProviderServer.bat
That's it.

Installer must be changed to :
1. Install PixiThumbnailProvider.dll from PixiEditor\Installer\PixiThumbnails and bat files to the app's folder
2. Issue the commands from bat files from app's dirctory. There shall be a silent option used /s -  Prevents displaying messages.

### DIFF
--- a/src/PixiEditor/Properties/AssemblyInfo.cs
+++ b/src/PixiEditor/Properties/AssemblyInfo.cs
@@ -50,5 +50,5 @@ using System.Windows;
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("0.1.9.7")]
-[assembly: AssemblyFileVersion("0.1.9.7")]
+[assembly: AssemblyVersion("0.1.9.8")]
+[assembly: AssemblyFileVersion("0.1.9.8")]


flabbet:
So technically, if there was C++ native code PixiParser implementation, no temporary image would be needed? All of the files would load, no matter if opened and saved again?

tomaszkot:
> So technically, if there was C++ native code PixiParser implementation, no temporary image would be needed? All of the files would load, no matter if opened and saved again?

No need to have any native code in Pixi. I meant extracting parts of the app into a .net dll. That would be a good move anyway and would allow to write a bridge from c++ to C# using e.g. CLI c++.

flabbet:
Ah, ofc. We are going to extract parts of the app into separate dlls, but PixiParser is already a separate dll

tomaszkot:
> Ah, ofc. We are going to extract parts of the app into separate dlls, but PixiParser is already a separate dll

OK, Then I would provide an alternative implementation of Thumbnails using PixiParser. I'd leave pure native (relying on temp save file) as alternative one.

flabbet:
Also, small PixiEditor icon in bottom right corner is required, otherwise people might confuse it with PNG or JPG 
image

tomaszkot:
> Also, small PixiEditor icon in bottom right corner is required, otherwise people might confuse it with PNG or JPG
> 
> This is how Affinity Photo does that image

Yeah, shall be easy to add...

tomaszkot:
image

Almost there!

tomaszkot:
Some considerations for the new Native+CLR approach vs prev native-only approach:
pros :
- Copy of a pixi file instantly gets proper thumbnail
- No additional save to png needed in PixiEditor

cons:
- Two more projects (CLR - native to .NETBridge, .NetThumbnailProvider - preparing thumbnail using PixiParser.Skia.dll)
- Could not debug CLR  and .NET projects. Not a big deal, and probably could be asked on SO, saw there hints for .net 4.0 but not for 6.0
- I have lot's of ugly warnings like : c:\windows\assembly\gac_64\system.data\2.0.0.0__b77a5c561934e089\system.data.dll : warning C4945: 'IDataRecord': cannot import symbol from 'c:\windows\assembly\gac_64\system.data\2.0.0.0__b77a5c561934e089\system.data.dll': as 'System::Data::IDataRecord' has already been imported from another assembly 'System.Data.Common'
Do not know how to fix it, maybe this is just my local pc problem.

flabbet:
Nice! Here is full res logo you can use.

PixiEditorLogo
d

tomaszkot:
I'd provide code soon, maybe tomorrow.  I was going to provide both options Native and Native_CLR_NET switchable in the PixiEditor config file, but I'm not sure this is needed from your point of view.

tomaszkot:
I think I have done it, yet on CI it shows a strange error :/ :
Project PixiCLRBridge is not compatible with native (native,Version=v0.0). Project PixiCLRBridge supports: net6.0-windows10.0.22000 (.NETCoreApp,Version=v6.0)

Equbuxu:
Just a note: we haven't forgotten about this PR, the current idea is to add a small pre-rendered preview png into .pixi and load it from C++. It will be added in the next PixiParser version. This way the C++ code will be way simpler, there will be no need for C++/C# interop and we won't need to update the rendering code as PixiEditor gets more advanced features, like blend modes, clip to layer below, etc.

tomaszkot:
> Just a note: we haven't forgotten about this PR, the current idea is to add a small pre-rendered preview png into .pixi and load it from C++. It will be added in the next PixiParser version. This way the C++ code will be way simpler, there will be no need for C++/C# interop and we won't need to update the rendering code as PixiEditor gets more advanced features, like blend modes, clip to layer below, etc.

Glad to hear it, let me know once the version with preview is ready.

Equbuxu:
@tomaszkot Hey, the .pixi file format was finally updated to include a pre-rendered preview, so if you want, you can finally continue working on this PR. Please let us know if are still interested in that. If you don't, that's okay, we will most likely merge your PR anyway and update the code ourselves.

tomaszkot:
OK, I'll have a look and let you know soon.

Issue #405, opened 2022-04-26
ViewModelMain.Current is null when the recovered documents are being loaded


Equbuxu:
```csharp
public void RecoverDocuments()
{
    MainWindow window = MainWindow.CreateWithDocuments(CrashReport.RecoverDocuments());
    
    Application.Current.MainWindow = window;
    window.Show();
    hasRecoveredDocuments = false;
}
```
This is the function that recovers the documents from a crash report. CrashReport.RecoverDocuments() creates Document objects. This happens before the main window and viewmodelmain are created. Document, however, expects the ViewModelMain to exist during construction and afterwards. This causes crashes in various places, like

```
   at PixiEditor.Models.DataHolders.Document.SetAsActiveOnClick(Object obj) in D:\a\1\s\PixiEditor\Models\DataHolders\Document\Document.cs:line 212
   at PixiEditor.Helpers.RelayCommand.Execute(Object parameter) in D:\a\1\s\PixiEditor\Helpers\RelayCommand.cs:line 40
   ...
```

```
   at PixiEditor.Models.DataHolders.Document.UpdateRecentlyOpened(String newPath) in D:\a\1\s\PixiEditor\Models\DataHolders\Document\Document.IO.cs:line 57
   at PixiEditor.Models.DataHolders.Document.set_DocumentFilePath(String value) in D:\a\1\s\PixiEditor\Models\DataHolders\Document\Document.IO.cs:line 20
   at PixiEditor.Models.DataHolders.Document.SaveWithDialog() in D:\a\1\s\PixiEditor\Models\DataHolders\Document\Document.IO.cs:line 40
   at PixiEditor.ViewModels.SubViewModels.Main.FileViewModel.SaveDocument(Object parameter) in D:\a\1\s\PixiEditor\ViewModels\SubViewModels\Main\FileViewModel.cs:line 252
   ...
```
etc.

Sometimes documents try to access the ViewModelMain right away, throwing exceptions that get caught in CrashReport.RecoverDocument(). Those documents are then discarded, even though there is nothing wrong with them

Equbuxu:
#393 is the same bug

Issue #294, opened 2021-12-17
Tool quick usage (using a tool while holding a button and then auto switching back) bug


Equbuxu:
Right now it is determined based on how long you hold the button down for but actually, it should switch back if you start using the tool before releasing the button

flabbet:
I'm not sure about that. This doesn't seem very convienient. If I tap a tool shortcut I expect it to switch to it, not wait for me to draw anything

Equbuxu:
> I'm not sure about that. This doesn't seem very convienient. If I tap a tool shortcut I expect it to switch to it, not wait for me to draw anything

I don't think you get it. Right now quick usage is determined by repeat events, it works like this:

Press E -> Eraser is selected -> Release right away (no repeat events) -> Eraser stays selected (you've just switched your tool)
Press E -> Erase for a bit (repeat events are firing) -> Release -> Old tool is selected (quick usage)
Press E -> Erase very quickly (no repeat events cause you're faster than them) -> Release -> Eraser stays selected (you wanted to "quick use" the eraser but you was a bit too quick and it switched instead)

Quick usage should instead be determined by whether you've started using the tool before or after releasing the E key:

Press E -> Eraser is selected -> Release without erasing -> Eraser stays selected (you've just switched your tool)
Press E -> Erase -> Release -> Old tool is selected (quick usage)
Press E -> Erase very quickly -> Release -> Old tool is selected (quick usage actually worked, cause it knows you erased something)

PR #414, opened 2022-07-05
Prototype integration

Equbuxu:
Important:
- [x] Clip canvas (Needs ChangeableDocument implementation)
- [x] Center content (Needs ChangeableDocument implementation)
- [x] Magic wand tool (Needs ChangeableDocument implementation)
- [x] Buttons and commands that flip viewport
- [x] Export image
- [x] Rectangle tool and rotate viewport tool have the same default shortcut "R"
- [x] #419
- [x] Enabling grid lines makes breaks zoom on small canvases
- [x] Grid lines look bad
- [x] Update channel doesn't allow to rollback
- [x] Brightness tool kinda sucks because I tried implementing it with skia
- [x] Reference layer, see https://github.com/PixiEditor/PixiEditorPrototype/issues/51
- [x] Pipette tool reference layer support
- [x] Line tool transform overlay
- [x] Rollback CPU repalce color solution and remove ComputeSharp
- [x] Make current shaders non dynamic (remove dynamic library and add proper attribute to current shaders)
- [x] Some images created in earlier versions of PixiEditor aren't imported correctly (will elaborate later, note to self: see drawing-mar6.pixi)
- [x] Fill tool works kinda like a pen tool when used outside selection (it shouldn't do anything)
- [x] PixiEditor crashes when trying to load an empty file (create new, save right away, close, open saved)

Important, but we can make a dev release before fixing these:
- [ ] Fix stuff from the review
- [ ] Extreme lag when holding down E to use the eraser and erasing. Edit: "Fixed" by limiting canvas update rate to 60fps. It's not a very good solution, but it's also not awful
- [ ] Lasso tool is generally laggy, probably because every single path change is propagated to the viewmodel, whereas just using the last one in the batch should be enough.
- [x] It's probably not possible to recover documents after a crash since `CreateWithDocuments` is commented out. Also, #405 shouldn't be too hard to fix by loading the documents after PixiEditor fully opens
- [ ] Make sure that the opened documends are saved during a backend crash
- [ ] `UpdateTaskbarIcon` is only called on setting change, so I'm assuming it's also broken
- [ ] Floodfill can "go around" the canvas borders when a chunk extends there (https://discord.com/channels/726445255175438346/726445255175438349/1043604357498814544)
- [ ] Fix resize canvas change such that it doesn't fail validation on startup (it should allow resizing to the same size)
- [ ] Look through all comments to make sure no important code is commented out
- [ ] Floodfill spills outside selection https://discord.com/channels/726445255175438346/726445255175438349/1044728599988682802

Nice to have:
- [x] Fix tests
- [ ] Fix all warnings
- [ ] Fix all binding errors
- [ ] Flip image
- [ ] Rotate image (90/180/270 degrees)
- [ ] What's new screen
- [ ] A way to apply mask (Needs ChangeableDocument implementation)
- [ ] Figure out how to make DocumentViewModel disposable
- [ ] Onboarding window
- [ ] .NET 7
- [ ] I kinda broke the feature where you can hold down a key to temporarily use a tool, so maybe we should finally implement it properly (#294)
- [ ] Dev Update channel should fetch from channels "above", like dev should check if release is newer
- [ ] Map photoshop shortcuts

Unsorted:
- [x] Implement stylus commands
- [x] Add properties for MouseXOnCanvas and MouseYOnCanvas to show on the bottom right
- [x] Palettes, Swatches
- [x] palette replace color
- [x] Document names
- [x] Layers select multiple
- [x] Layers context menu
- [x] Delete layer/folder
- [x] Apply transform button and shortcut
- [x] You can't draw on leftmost and topmost pixels
- [x] EventToCommand -> InvokeCommandAction, remove mvvmlight
- [x] Discord
- [x] blend modes dropdown is ugly, and other layers UI too
- [x] Something is wrong with selection
- [x] Opacity text field lets you set opacity above 100%
- [x] Saving, Loading .pixi files
- [x] Add new layer stuff into PixiParser (Masks, Clipping etc.) (aka pixi 4.0)
- [x] Clipboard
- [x] Make canvas context menu work
- [x] Rotate canvas tool
- [x] Move tool: should probably enable transform selected area
- [x] Lasso tool
- [x] Fill tool (Make it work, don't forget to add an option to consider all layers since it's already implemented)
- [x] A proper system for transform overlay freedoms: shift key should lock proportions, rectangles should be rotateable, stuff like that
- [x] Allow rotation of transform overlay when dragging outside canvas, maybe also show a rotation cursor
- [x] Transform overlay can be squished into a line which makes it behave weirdly
- [x] Transform overlay rotation snapping
- [x] Highlight for tools: should be implemented as another overlay, maybe somewhat similar to how selection works
- [x] Pen tool doesn't let you draw single pixels when pixel perfect is enabled
- [x] A loading... window that pops up when the background thread takes it's time (it should simply be shown when DocumentViewModel.Busy is true)
- [x] Drag and drop bug: Make the layers window float, create a bunch of layers, grab the bottom right corner and resize it to be very small, then grab again and resize to original size. This causes PixiEditor to get stuck in a Drag-n-Drop state. The only way to get out of it is to press Esc (or kill the app).
- [x] Pen tool draws ugly circles, sizes 2 and 3 look the same, 4 isn't circular at all, etc.
- [x] Symmetry overlay and commands
- [x] Reset viewport position button
- [x] A text that shows current viewport rotation
- [x] Map aseprite shortcuts
- [x] Shortcuts overhaul
- [x] About window
- [x] Actually "holding down the key temporarily" no longer works because my brand new keyboard filter seems to block repeat events, this needs to be fixed.
- [x] A single click with the selection tool should clear selection
- [x] A create mask button


src/PixiEditor.ChangeableDocument/Changes/Drawing/ShiftLayer_UpdateableChange.cs
```diff
@@ -28,48 +43,57 @@ public void Update(VecI delta, bool keepOriginal)
...
-        return new LayerImageChunks_ChangeInfo(layerGuid, chunks);
+        _tempChanges.Clear();
+
+        foreach (var layerGuid in layerGuids)
+        {
+            var chunks = ShiftLayerHelper.DrawShiftedLayer(target, layerGuid, keepOriginal, delta);
+            _tempChanges.Add(new LayerImageChunks_ChangeInfo(layerGuid, chunks));
+        }
+        
+        return _tempChanges;
```
Equbuxu, line 75: looks sus, why keep a reference to a returned list locally? who knows what it's gonna be used for by the caller. Also why store it at all?

src/PixiEditor.ChangeableDocument/Changes/Root/CenterContent_Change.cs
```diff
@@ -0,0 +1,109 @@
...
+    public CenterContent_Change(List<Guid> layers)
+    {
+        _affectedLayers = layers;
+    }
+    
+    public override bool InitializeAndValidate(Document target)
+    {
+        if (_affectedLayers.Count == 0)
+        {
+            return false;
```
Equbuxu, line 22: a tip: Returning false is very similar to throwing an exception, in a sense that it should never happen. So ideally we should only return false when something clearly doesn't make sense. I can see how pressing center content on a document with 0 layers would result in a validation failure here. So maybe we should handle it as a valid case (by doing nothing and setting ignoreInUndo to true)? tbh not sure, so this is just a tip

Edit: I see now that it the UI handles the "no layers selected" case so we can keep it as is

src/PixiEditor/ViewModels/SubViewModels/Document/DocumentManagerViewModel.cs
```diff
@@ -50,7 +50,10 @@ public DocumentManagerViewModel(ViewModelMain owner) : base(owner)
     [Command.Basic("PixiEditor.Document.ClipCanvas", "Clip Canvas", "Clip Canvas", CanExecute = "PixiEditor.HasDocument")]
     public void ClipCanvas()
     {
-        //Owner.BitmapManager.ActiveDocument?.ClipCanvas();
+        if (ActiveDocument is null)
+            return;
+        
+        ActiveDocument?.Operations.ClipCanvas();
```
Equbuxu, line 56: gotta be extra sure that activedocument isn't null

flabbet (approved):
minor stuff, looks about right

Issue #412, opened 2022-07-04
Text box select bug


aventurain:
**When it happens**

It happens in Pallet color selection window
![image](https://user-images.githubusercontent.com/64274465/177212001-f5f69861-3cab-4eef-9ff8-28ba640e96dc.png)
and Settings window
![image](https://user-images.githubusercontent.com/64274465/177212058-2df816bd-7525-456c-a792-2fedeb144014.png)

window will minimize when you follow next steps: 

**Reproduction steps**

In Pallet color selection window:
If you release the mouse button outside the window during the text selection

In Settings window:
Bug happening when you select text and click in window, but outside text box.

**Expected**
The window should not minimize

Bayrem00:
The color map loss of focus bug is part of the ColorPicker code and can not be fixed on this repo.

Equbuxu:
related #418

PR #420, opened 2022-07-17
Fix focus loss if size text is selected in settings view

Bayrem00:
Fix for  [#412](https://github.com/PixiEditor/PixiEditor/issues/412)

Focus the next view instead of main in size text boxes.
The color map loss of focus bug is part of the ColorPicker code and can not be fixed on this repo.


Equbuxu:
Seems okay as a temporary fix, but the real underlying issue is that `TextBoxFocusBehavior`'s `RemoveFocus` function makes the assumption that `MainWindow.Current.mainGrid` is an element that always exists and is always somewhere near the root of the visual tree. This isn't the case for windows that aren't the main one. `FocusNext` seems to work fine for our current UI layout, but in some cases it can cause weird behavior, say you could be trying to remove focus from a textbox, and it would instead focus the next textbox.

Equbuxu:
Implemented a better fix in 930c677d8022420113b2d01d379f1cea8a8a8db5

PR #424, opened 2022-10-08
Drawing API Core

flabbet:
A drawing backend API that separates pure logic of Skia from PixiEditor


Equbuxu:
- Using the selection tool causes a crash
- Locking layer transparency doesn't do anything (works in the prototype integration branch)

Equbuxu (changes requested):
forgot

flabbet:
fixed, try me again

PR #430, opened 2022-11-28
Upgraded to .NET 7

flabbet:


Equbuxu (approved):
haven't looked, but looks good

PR #433, opened 2022-11-30
random ui stuff

Equbuxu:
- [x] Don't show transform controls when transforming with the cursor tool
- [x] Transform overlay no longer jumps because of rounding (#436)
- [x] Fix shortcut textboxes in settings getting keyboard focus when the settings window becomes active (like if you switch to a different window and then back)
- [x] Move selection to the new layer when you create one
- [x] Fix tool change applying transform for move/rotate viewport tools (#436)
- [x] Fix transform selected area change failing validation when the selection is fully outside canvas
- [x] Holding down shift and scaling down the selected area to a single pixel leads to a crash


flabbet:
Hmm, hiding apply transform in cursor tool is questionable in current form for me. Since after you move, you can't undo unless you switch tool and use it. It's kinda frustrating

flabbet (changes requested):
Overall, everything looks good. But unless we discuss cursor tool undo thing, changes requested

flabbet (approved):

Issue #342, opened 2022-02-09
Add a warning when saving into formats other than .pixi


Equbuxu:
If you open a .png image, add a new layer to it and then try to save it, the layers will be automatically merged and saved into a png. This is all good, but I think people might not realize that the layer information is being lost even though they've saved the file.

Equbuxu:
Still not fully resolved, hints are good, but you can still hit ctrl+s and it will behave as it did before

PR #438, opened 2022-12-26
Added examples to save and export menu items

Tech-FZ:
## What I added

Update: The export warning has now been removed and the Save and Export menu items now give examples for file types.

ApplicationFrameHost_Vy4G9PCG6w

The idea was requested via this issue by another user: https://github.com/PixiEditor/PixiEditor/issues/342


src/PixiEditor/Models/Dialogs/ExportFileDialog.cs
```diff
@@ -78,11 +78,19 @@ public override bool ShowDialog()
...
+
         if (popup.DialogResult == true)
         {
             FileWidth = popup.SaveWidth;
             FileHeight = popup.SaveHeight;
             FilePath = popup.SavePath;
+
+            if (ChosenFormat.ToString() != "pixi")
+            {
+                ExportWarningPopup exportWarningPopup = new ExportWarningPopup();
```
flabbet, line 90: This is not a good solution. PixiEditor already has dialog popups that you can customize.

src/PixiEditor/Models/Dialogs/ExportFileDialog.cs
```diff
@@ -78,11 +78,19 @@ public override bool ShowDialog()
...
         ExportFilePopup popup = new ExportFilePopup(FileWidth, FileHeight);
         popup.ShowDialog();
+
         if (popup.DialogResult == true)
         {
             FileWidth = popup.SaveWidth;
             FileHeight = popup.SaveHeight;
             FilePath = popup.SavePath;
+
+            if (ChosenFormat.ToString() != "pixi")
```
flabbet, line 88: Displaying popup each time user wants to export will be very annoying. This should be saved in editor data (user closed popup) so it only shows once.

flabbet (changes requested):
I think this can be solved much easier. Most users know what .png file format is. So instead of popup explaining what export does, I'd simply replace "Export..." with "Export... (ex. .png, .jpg) and "Save" with "Save (.pixi). 

Another good solution would be to add a little question mark next to the text in context menu item with tooltip what the option does. Example: [Save (?)] where '?' is an icon

Equbuxu:
> I think this can be solved much easier. Most users know what .png file format is. So instead of popup explaining what export does, I'd simply replace "Export..." with "Export... (ex. .png, .jpg) and "Save" with "Save (.pixi).
> 
> Another good solution would be to add a little question mark next to the text in context menu item with tooltip what the option does. Example: [Save (?)] where '?' is an icon

I have some comments on that, so let's discuss on discord first. As for the tests, we don't really have any tests for the UI right now, so don't worry about it.

src/PixiEditor/Views/MainWindow.xaml
```diff
@@ -178,13 +178,13 @@
...
-                            Header="_Save"
+                            Header="_Save (.pixi)"
                             cmds:Menu.Command="PixiEditor.File.Save" />
                         <MenuItem
-                            Header="_Save As..."
+                            Header="_Save As... (.pixi)"
                             cmds:Menu.Command="PixiEditor.File.SaveAsNew" />
                         <MenuItem
-                            Header="_Export"
+                            Header="_Export (.png, .jpeg, .gif, .bmp)"
```
flabbet, line 187: This is a bit too long, I'd replace it with "_Export (.png etc.)"

flabbet (changes requested):

src/PixiEditor/Views/MainWindow.xaml
Equbuxu: ".png etc." sounds broken, maybe ".png, .jpg, etc."

flabbet: ok, ".png, .jpg, etc" then

flabbet (approved):

PR #439, opened 2022-12-27
Mouse Move update limiting hacky fix

flabbet:
so generally, right now MouseMove event should be handled by MouseUpdateController, which periodically unsubscribes and subscribes from MouseMove event, so mouse updates are limited to given amount. This is better than discarding updates using code and some counter, because it doesn't invoke underlying WPF code that many times.


Equbuxu (approved):
looks good, it's insane how well this works

PR #452, opened 2023-01-29
Parser support for clip and blend mode fixes

### LINKED ISSUE #447: Blend modes and clip to layer below is not saved in .pixi
Change blend mode for layer, clip another layer onto some other layer and save it as .pixi. Load file, all this info is lost

flabbet:
#447
https://github.com/PixiEditor/PixiParser/pull/19


Equbuxu (approved):
looks good to me, you can also try fixing #454 in this pr

Equbuxu (approved):
works well, there is this issue #457 that is also related
feel free to ignore it tho if you don't feel like fixing it now, we can leave it for some patch after the release

Issue #443, opened 2023-01-20
Separate undo for a transform rect


flabbet:
Currently, after moving a transform using move tool with selected area, you can't undo or redo your transformation unless you switch and use other tool. Very frustrating, each transform move should be recorded in undo.

PR #455, opened 2023-01-31
Enhance transform overlays

### LINKED ISSUE #445: Selection tools should look at shift and alt to determine what selection type to use
No key -> Normal
Shift -> Add
Alt -> Subtract

### LINKED ISSUE #446: Line tool overlay get created in the wrong place when holding shift
Draw a horizontal line by holding down shift. Once you finish, the second point of the overlay is created under your cursor and not where the line ends

Equbuxu:
- [x] Implement undo/redo for line tool transform overlay
- [x] Implement undo/redo for the main transform overlay (#443)
- [x] Fix #446
- [x] Implement nudging the transform by 1 pixel using arrow keys
- [x] Fix #445


flabbet (changes requested):
Add reference layer -> Click arrow down -> weird black pixels (wpf element selection)

Also I wasn't sure about meging nudge undo, but after playing with it for a while, it seems like good solution.

flabbet (approved):

Issue #458, opened 2023-02-07
Build without UPDATE flag still shows update settings


flabbet:
Repro:

1. Switch to Dev Steam solution configuration
2. Build and run
3. Go to settings -> General

There shouldn't be update settings, since Dev Steam configuration doesn't have UPDATE compiler flag, however this case seems not to be handled.

PR #459, opened 2023-02-09
Handled update settings for UPDATE and non update platforms

flabbet:
Steam and MSIX configuration now shows in settings "Handled by {platform}" in update stream and update checkbox is grayed out. Also updated MSIX project target platform 

#458


CPKreuz:
why not grey out the combox box?

flabbet:
I could do that, but then I'd have to make ComboBox very wide so it contains full "Handled by Microsoft Store" text. If it's not grayed out, then user can click on that and will see full name. It's of course fixable, but imo too much effort for almost the same effect

flabbet:
Done

flabbet:
Also hidden donation button for Steam

CPKreuz (approved):
Looks good

PR #465, opened 2023-02-16
Color replacer replaces color in palette too

flabbet:
And swaps color in replacer, so user can quickly revert

### DIFF
--- a/src/PixiEditor/Models/DocumentModels/Public/DocumentOperationsModule.cs
+++ b/src/PixiEditor/Models/DocumentModels/Public/DocumentOperationsModule.cs
@@ -234,7 +234,19 @@ internal class DocumentOperationsModule
     {
         if (Internals.ChangeController.IsChangeActive || oldColor == newColor)
             return;
+        
         Internals.ActionAccumulator.AddFinishedActions(new ReplaceColor_Action(oldColor, newColor));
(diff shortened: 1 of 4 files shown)


src/PixiEditor/Models/DocumentModels/Public/DocumentOperationsModule.cs
```diff
@@ -234,7 +234,19 @@ public void ReplaceColor(Color oldColor, Color newColor)
...
+    }
+
+    private void ReplaceInPalette(Color oldColor, Color newColor)
+    {
+        int indexOfOldColor = Document.Palette.IndexOf(oldColor);
+        if(indexOfOldColor == -1)
+            return;
+        
+        Document.Palette.RemoveAt(indexOfOldColor);
+        Document.Palette.Insert(indexOfOldColor, newColor);
```
CPKreuz, line 249: A simple `Document.Palette[indexOfOldColor] = newColor;` should have done the job too

CPKreuz (approved):
Looking good

src/PixiEditor/Models/DocumentModels/Public/DocumentOperationsModule.cs
flabbet: oh, right lmao

PR #471, opened 2023-02-22
Cropped layer previews

Equbuxu:


Equbuxu:
I don't think anyone will understand the code, but please test very thoroughly (main previews, mask previews, layers, folders, etc.), I had so many crashes there is no way it works 100% now

flabbet (changes requested):
Open an image from disk -> Select whole image pixels -> scale it down: preview bitmap is not cropped and won't be.

flabbet:
Resizing Canvas causes a crash

flabbet (changes requested):
Create a 64x64 canvas -> Draw something -> Resize canvas to 128x128 -> Undo.

Preview is weirdly duplicated

PR #472, opened 2023-02-22
Separate thread for the global mouse hook

Equbuxu:
Moved global mouse hook into a separate thread. The cursor doesn't lag on startup anymore.


src/PixiEditor/Helpers/LowLevelWindow.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private void Dispose(bool disposing)
+    {
+        if (!disposed)
+        {
+            if (disposing)
+            {
+                // Dispose managed resources
+            }
```
flabbet, line 92: uhhh

flabbet (approved):

Issue #476, opened 2023-02-24
The reference layer does not fully retain it's position after save and reload


Equbuxu:
https://user-images.githubusercontent.com/29402050/221284622-b65a0e20-e662-4588-a94b-1ebd45085bb8.mp4

PR #482, opened 2023-03-01
Fixed reference layer shape serialization and recent file not found

flabbet:
#476 and added file not found to recently opened file

### DIFF
--- a/src/PixiEditor/Helpers/DocumentViewModelBuilder.cs
+++ b/src/PixiEditor/Helpers/DocumentViewModelBuilder.cs
@@ -5,6 +5,7 @@ using ChunkyImageLib.DataHolders;
 using PixiEditor.DrawingApi.Core.ColorsImpl;
 using PixiEditor.DrawingApi.Core.Numerics;
 using PixiEditor.DrawingApi.Core.Surface;
+using PixiEditor.Parser;
 using BlendMode = PixiEditor.ChangeableDocument.Enums.BlendMode;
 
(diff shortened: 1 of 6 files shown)


CPKreuz (approved):
Looks good

src/PixiEditor/Models/Enums/ExecutorType.cs
```diff
@@ -5,3 +5,9 @@ internal enum ExecutorType
     Regular,
     ToolLinked,
 }
+
+internal enum ExecutorStartMode
+{
+    OnSelectTool,
```
Equbuxu, line 11: maybe something like "RightAway", since usually executors start right away

src/PixiEditor/ViewModels/SubViewModels/Tools/ToolViewModel.cs
```diff
@@ -53,7 +53,7 @@ public bool IsActive
 
     public Toolbar Toolbar { get; set; } = new EmptyToolbar();
 
-    public virtual void UpdateActionDisplay(bool ctrlIsDown, bool shiftIsDown, bool altIsDown) { }
+    public virtual void ModiferKeyChanged(bool ctrlIsDown, bool shiftIsDown, bool altIsDown) { }
```
Equbuxu, line 56: Modifier*

Equbuxu (changes requested):

Equbuxu (approved):
haven't tried launching it so imma trust you that it works

PR #499, opened 2023-04-05
Localization

flabbet:
Added localization support.

Included languages: English, Polish


Equbuxu:
image

Equbuxu:
EDITING_IMG key is not present in the code, should be used in discord preview

flabbet:
It can be removed, we decided not to localize Discord Rich Presence

Equbuxu:
- Change your language and then navigate to Keybindings > Tools. String interpolation doesn't work, all tools have the same label
- "RENAME" 
image
- interest instead of intersect
image
- not sure if this is the right PR, but this looks confusing:
image
- not sure if this is the right PR, but what has happened here:
image
- from what I remember we decided against localizing filenames 
image
- pen tool action display doesn't show the right thing, it should be PEN_TOOL_ACTION_DISPLAY
image
- seems like CLICK_SELECT_PRIMARY isn't visible anywhere despite being in the code (or I'm just dumb)

flabbet:
> seems like CLICK_SELECT_PRIMARY isn't visible anywhere despite being in the code (or I'm just dumb)
It is in viewport context menu when having palette and hovering over a color

Equbuxu:
Discord rich presence is still localized

FILE_NOT_FOUND_PATH_FULL_PATH is still only present in POEditor (I have tried searching with visual studio, rider, and grep)

Version: ... text in about popup is still not localized

image

might still find more stuff, I need to look through the remaining keys

flabbet:
Memo: Localize Drop Palette Here

flabbet:
review pls

Equbuxu (changes requested):
- not localized
image

- CLIP_TO_BELOW and CLIP_LAYER_BELOW should be combined into a single key
- can't seem to find these keys in code, also they have a space in them
image

PR #500, opened 2023-04-06
Fixed export filename crash

flabbet:
For some reason, file dialog doesn't append extension for filename like "100.001" so PixiEditor couldn't detect extension. Simple fallback works here

### DIFF
--- a/src/PixiEditor/Helpers/SupportedFilesHelper.cs
+++ b/src/PixiEditor/Helpers/SupportedFilesHelper.cs
@@ -36,6 +36,27 @@ internal class SupportedFilesHelper
         return allFileTypeDialogsData.Where(i => i.FileType == type).Single();
     }
 
+    public static string FixFileExtension(string pathWithOrWithoutExtension, FileType requestedType)
+    {
+        if (requestedType == FileType.Unset)
+            throw new ArgumentException("A valid filetype is required", nameof(requestedType));
+
+        var typeFromPath = SupportedFilesHelper.ParseImageFormat(Path.GetExtension(pathWithOrWithoutExtension));
+        if (typeFromPath != FileType.Unset && typeFromPath == requestedType)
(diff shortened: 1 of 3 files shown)


Equbuxu:
This is not a proper review, but I suspect that this solution won't be ideal, as even though the savefiledialog doesn't append the extension for whatever reason, it doesn't mean the user haven't chosen one. So I suspect it will ignore user's choise and default to png instead. Also, we had an issue with the filename becoming too long after the extension is appended.

We already have a saving method that tries to handle a lot of the weirdness of SaveFileDialog, it's Exporter.TrySaveWithDialog(), I think it would be best to improve it to support different file formats and images, and use it everywhere where we need to save something

flabbet:
I jest tested it with paint.net, intrestingly, using 100.001 as a file name, causes paint.net to save the file as a 001 archive. This seems like a Windows bug or weirdiness? 
image

Equbuxu:
> I jest tested it with paint.net, intrestingly, using 100.001 as a file name, causes paint.net to save the file as a 001 archive. This seems like a Windows bug or weirdiness? Since Paint.net falled back to some unexpected result image
> 
> UPDATE: The same happens with regular Paint and Affinity Photo

are you sure the "archive" isn't just your image with a wrong file extension?

CPKreuz:
The SaveFileDialog has a FilterIndex property, idk if it actually sets the index the user set but maybe that could work

https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.filedialog.filterindex?view=windowsdesktop-7.0#system-windows-forms-filedialog-filterindex

Equbuxu:
The function in Exporter literally already handles all of this, as all these issues aren't new, they were already fixed for the "save as" option. The function just needs to be adaptes slightly to work for Export

flabbet:
Fixed

Equbuxu (approved):

Issue #498, opened 2023-04-04
Differentiate between exception types in Exporter.TrySaveAs


Equbuxu:
Exporter.TrySaveAs returns false for any exception. Might be good to catch some specific exceptions and return different SaveResult's based on them. Note that DialogSaveResult should contain everything SaveResult contains, and that their usages also need to be updated (there is at least one switch that switches on DialogSaveResult)

Issue #495, opened 2023-04-01
Environment.SpecialFolder.LocalApplicationData might be different from %LocalAppData% when running from ms store


Equbuxu:
This is based on a conversation with @Krist0fer#3857 (<@757991981904429167>) on discord. They were using the ms store version of PixiEditor, and it mostly worked fine (settings and recent files were saved properly anyway), but for whatever reason the %LocalAppData%\PixiEditor folder was completely missing. Interestingly, it's the only place in the app where "%LocalAppData%" was used, the rest of the app uses Environment.SpecialFolder.LocalApplicationData to find the path.

@Krist0fer#3857 later downloaded the steam version of PixiEditor, after which the %LocalAppData%\PixiEditor directory with all the usual stuff was created. So it must be an issue related specifically to the ms-store

Potential solution: just use Environment.SpecialFolder.LocalApplicationData everywhere instead of %LocalAppData%

PR #504, opened 2023-04-07
Random improvements

Equbuxu:
- [x] Crash reports now state whether the build came from steam, website, or msstore
- [x] #498 
- [x] #495
- [x] Catch "System.ComponentModel.Win32Exception: An error occurred trying to start process 'https://pixieditor.net' with working directory 'C:\Program Files (x86)\Steam\steamapps\common\PixiEditor'. Application not found"
- [x] Added a helper that send a stacktrace to the webhook in situations other that crashes. The idea is that in some cases it can be completely safe to catch all exceptions and just show and error to the user, but at the same time we'd wanna know what these exceptions are, because they can still be cause by our own bug. So in this case we can log the exception using the helper method.
- [x] Don't throw when trying to create surfaces larger than 10k x 10k
- [x] The recent file list doesn't show missing files anymore

### DIFF
--- a/src/PixiEditor/Helpers/VersionHelpers.cs
+++ b/src/PixiEditor/Helpers/VersionHelpers.cs
@@ -14,77 +14,27 @@ internal static class VersionHelpers
     {
         StringBuilder builder = new(GetCurrentAssemblyVersion().ToString());
 
-        bool isDone = false;
-
-        AppendDevBuild(builder, ref isDone);
-        if (isDone)
-            return builder.ToString();
-
-        AppendMsixDebugBuild(builder, ref isDone);
-        if (isDone)
-            return builder.ToString();
-
-        AppendDebugBuild(builder, ref isDone);
-        if (isDone || !moreSpecific)
-            return builder.ToString();
-
-        AppendSteamBuild(builder, ref isDone);
-        if (isDone)
-            return builder.ToString();
-
-        AppendMsixBuild(builder, ref isDone);
-        if (isDone)
-            return builder.ToString();
-
-        AppendReleaseBuild(builder, ref isDone);
-        return builder.ToString();
-    }
-
-    [Conditional("DEVRELEASE")]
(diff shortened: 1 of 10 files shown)


src/PixiEditor/Helpers/VersionHelpers.cs
```diff
@@ -43,4 +55,36 @@ private static void AppendDebugBuild(StringBuilder builder, ref bool done)
...
+        builder.Append(" MSIX Build");
+    }
+
+    [Conditional("STEAM")]
+    private static void AppendSteamBuild(StringBuilder builder, ref bool done)
+    {
+        done = true;
+
+        builder.Append(" Steam Build");
+    }
```
flabbet, line 89: Wouldn't it be simpler to just call one method and put preprocessors inside:

```csharp
private static void AppendBuild(StringBuilder builder)
{
    string str = "";
#if STEAM
    str = " Steam Build"
#elif MSIX
   str = " MSIX Build"

...etc
}
```

src/PixiEditor/ViewModels/SubViewModels/Main/MiscViewModel.cs
```diff
@@ -21,6 +22,14 @@ public MiscViewModel(ViewModelMain owner)
...
     {
-        ProcessHelpers.ShellExecute(url);
+        try
+        {
+            ProcessHelpers.ShellExecute(url);
+        }
+        catch (Exception e)
+        {
+            NoticeDialog.Show(title: "Error", message: $"Couldn't open the address {url} in your default browser");
+            _ = CrashHelper.SendExceptionInfoToWebhook(e);
```
flabbet, line 32: shouldn't it be async? What if someone has poor internet connection? Will it just hang?

flabbet: I'd make async command implementation for that

flabbet (changes requested):

flabbet (approved):

PR #515, opened 2023-04-16
New palette formats parsers

flabbet:
- Added png palette parsing
- Added Paint.net txt file parser
- Added Hex file parser
- Added Gimp GPL palette file parser
- Updated some Nuget packages
- #525


CPKreuz (changes requested):
* When trying to drop a png file, it also opens the file open dialog
* When opening a big png file, maybe ask the user if they are sure they want to import a palette or are just trying to drop in a image
* I was not able to drop in any of GIMP's built in GPL file's. I guess they are using a old format? here's one of them: [Warm_Colors.txt](https://github.com/PixiEditor/PixiEditor/files/11302526/Warm_Colors.txt)
* I think it would be cool to be able to open .pixi files as palettes

flabbet:
@CPKreuz I addressed issues you mentioned, good catches. But I ignored big png file

flabbet:
Also I added discard current palette

CPKreuz (changes requested):
Maybe disable these when there's no color in the palette

image

src/PixiEditor/Data/Localization/Languages/en.json
```diff
@@ -426,6 +426,8 @@
     "BROWSE_PALETTES": "Browse palettes",
     "LOAD_PALETTE": "Load palette",
     "SAVE_PALETTE": "Save palette",
+    "DISCARD_PALETTE": "Discard palette",
+    "DISCARD_PALETTE_CONFIRMATION": "Are you sure you want to discard current palette? This cannot be undone.",
```
CPKreuz, line 430: I think this sounds better

Are you sure you want to discard *__the__* current palette? This cannot be undone.

CPKreuz (approved):
Looks good, maybe a small localization change tho

src/PixiEditor/Data/Localization/Languages/en.json
flabbet: For me it sounds wierd, however this can be changed in POEditor

PR #530, opened 2023-05-11
Color Picker, localization and transform improvements

CPKreuz:
* Don't hide the canvas when there's no reference layer or the reference layer is outside the document bounds
* Don't hide the canvas or reference while there's a transform going on
* Added action displays while transforming
* Localized the Title property of the Window class for most popups
* Fixed missing `Localization Debug Window` title localization
* Changed DialogTitleBar.TitleText property to TitleKey
* Added a shortcut hint to the Apply Transform button


CPKreuz:
👀

Issue #522, opened 2023-04-22
Default mode on brightness tool works the same as repeat


flabbet:
bruh

PR #532, opened 2023-05-12
Fixed default mode for brightness tool

flabbet:
#522

### DIFF
--- a/src/ChunkyImageLib/ChunkyImage.cs
+++ b/src/ChunkyImageLib/ChunkyImage.cs
@@ -1089,7 +1089,7 @@ public class ChunkyImage : IReadOnlyChunkyImage, IDisposable
         affectedAreaSize = affectedAreaSize * scale;
         targetChunk.Surface.DrawingSurface.Canvas.ClipRect(new RectD(affectedAreaPos, affectedAreaSize));
 
-        operation.DrawOnChunk(targetChunk, chunkPos);
+        operation.DrawOnChunk(targetChunk, chunkPos, this);
         targetChunk.Surface.DrawingSurface.Canvas.RestoreToCount(count);
     }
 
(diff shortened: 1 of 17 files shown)


Equbuxu:
Injecting caller ChunkyImage into Operations is dangerous, at the very least you can get infinite recursion if you call anything non-committed, and I feel like there can be other more subtle issues too but don't quite remember what they can be

You could maybe inject the committed chunk but it doesn't seem like a good idea, operations aren't supposed to be aware of how internals of chunkyimage work. If you draw many things on a chunky image one after the other and call CommitChanges periodically, the precise moments of when you call of CommitChanges shouldn't affect the final result you get in any way, it doesn't make sense for CommitChanges to affect drawing, it's just for saving

Besides, the default mode of the brightness tool worked fine in 1.0.0.0 without any of these changes

Equbuxu (changes requested):
please check if we can implement the brightness tool like in 1.0.0.0

flabbet:
I fixed it differently, just by passing Func that gets commited pixel to draw operation, this seems like okay solution, not ideal, but at least it doesn't involve passing whole ChunkyImage and hacking ChangeBrightness_UpdateableChange with multiple temp surfaces and calculation of symmetry positions. Also my solution is more readable and flexible, a little more generic too

Equbuxu (changes requested):
soo the default mode of the brightness tool works like a white brush now
Edit: nevermind, it's not a white brush, it's more like a stamp tool? I guess it doesn't look at the right portion of the image
Edit2: 

https://github.com/PixiEditor/PixiEditor/assets/29402050/d16b88a9-c796-4838-b463-2736cfd189a4

flabbet:
wtf

Issue #488, opened 2023-03-09
Crop to selection


Equbuxu:
A feature to crop canvas to the tight bounds of the current selection

PR #537, opened 2023-05-27
Right click mode and other improvements

CPKreuz:
* Added crop to selection command #488 
* Changed default export shortcut to Ctrl + E
* Updated Discord's background color in HelloTherePopup to the current blurple
* Fixed open appdata commands
* Added right click mode

The implementation for right click mode is probably not the best and buggy when bringing the mouse cursor out of the viewport while drawing


flabbet:
image

src/PixiEditor/Models/DocumentModels/Public/DocumentOperationsModule.cs
```diff
@@ -564,6 +564,25 @@ public void SelectionToMask(SelectionMode mode)
         Internals.ActionAccumulator.AddFinishedActions(new SelectionToMask_Action(member.GuidValue, mode));
     }
 
+    public void CropToSelection(bool clearSelection = true)
+    {
+        if (Document.SelectionPathBindable.IsEmpty)
```
Equbuxu, line 569: would be nice to check for TightBounds having W > 0 and H > 0 too, so that the change doesn't fail validation

src/PixiEditor/Views/MainWindow.xaml.cs
```diff
@@ -247,4 +249,13 @@ private void MainWindow_OnKeyDown(object sender, KeyEventArgs e)
             e.Handled = true;
         }
     }
+
+    private void Viewport_OnContextMenuOpening(object sender, ContextMenuEventArgs e)
+    {
+        if (DataContext.ToolsSubViewModel.RightClickMode != RightClickMode.ContextMenu &&
+            DataContext.ToolsSubViewModel.ActiveTool is ShapeTool)
```
Equbuxu, line 256: I'd argue that the context menu should be fully disabled on other modes, instead of it being based on the tool (open for discussion tho)

Equbuxu (dismissed):

flabbet (changes requested):
Erase mode should erase in brightness tool
Selection tools should always have context menu, brightness tool should darken on right click with secondary color mode

flabbet (changes requested):
There shouldn't be a tool that doesn't have right click action. Zoom, rotate, move and eraser tool should also have context menu for right click for secondary color mode. On Erase mode, they should erase (I guess it also makes sense for Eraser, for the sake of muscle memory).

Second thing is no space between label and dropdown 
image

flabbet (changes requested):
No hover in dropdown 
image

Brightness tool opens context menu and darkens on secondary color mode

Color picker opens context menu on erase mode
Eraser opens context menu on erase mode (should erase on both)

flabbet (changes requested):
(Use erase mode) Select Rectangle Tool -> Select Eraser tool -> Erase with right click

Expected behavior: Eraser tool stays selected
Actual behavior: Selected tool gets switched back to rectangle tool.

Except that everything else seems good

src/PixiEditor.ChangeableDocument/Changes/Selection/SelectEllipse_UpdateableChange.cs
```diff
@@ -26,6 +27,8 @@ public void Update(RectI borders)
     public override bool InitializeAndValidate(Document target)
     {
         originalPath = new VectorPath(target.Selection.SelectionPath);
+        documentConstraint = new VectorPath();
+        documentConstraint.AddRect(new RectI(VecI.Zero, target.Size));
```
flabbet, line 31: You can easily omit this constraint by resizing canvas, I doubt it's a problem though. It applies to all constraints you made

flabbet (changes requested):

src/PixiEditor/Views/UserControls/Layers/FolderControl.xaml
```diff
@@ -93,6 +93,12 @@
                             </Grid>
                         </Border>
                         <StackPanel Orientation="Vertical"  Margin="3,0,5,0">
+                            <userControls1:EditableTextBlock
+                                x:Name="editableTextBlock"
+                                d:Text="New Folder" FontSize="14"
```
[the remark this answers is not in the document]
CPKreuz: The EditableTextBlock's both have 14

The opacity/blending mode text are actually inconsistent

flabbet (approved):
😘

Issue #540, opened 2023-06-02
Hex palette saving format is inconsistent


flabbet:
Saving a palette in .hex format causes common colors "letters" to compress in length, which is technically a correct hex format, however this can cause bugs in some programs, since they expect 6 characters long hex, including PixiEditor.

`HexPalettesParser` line `string hex = $"{color.R:X}{color.G:X}{color.B:X}";` in some cases R, G or B can be one letter long (F instead of FF)

Both our parser should handle this case and also save in 6 character long format

PR #541, opened 2023-06-02
Fixed .hex parser saving colors in bad hex format

flabbet:
#540


CPKreuz (approved):
Haven't tested but it looks like it should be working. Code also looks quite clean but more comments would be really helpful next time

PR #550, opened 2023-06-17
minor changes

flabbet:
I don't actually expect all of you to review everything, but try checking if you can crash palette browser and test if supporter pack works for you


CPKreuz (changes requested):
Localization not working

image

When hovering the image the title should become white too and use CultureInfo.CurrentCulture with the 'd' date format

flabbet:
> 1. Localization not working

Because localization file is in extension, and you don't have it installed

> 2. When hovering the image the title should become white too and use CultureInfo.CurrentCulture with the 'd' date format

I use 'd' date format in XAML

> 3. Also the window sizing should be adjusted so that the recent documents don't scroll

Sure

CPKreuz:
> I use 'd' date format in XAML

but not the right culture

flabbet:
Fixed most of the issues (besides white link text on image hover, because I got too frustrated while implementing that xd)

CPKreuz (changes requested):
Using a unkown news type breaks the fetching.
Also I don't think any new users installing PixiEditor should be bombarded with news. Maybe keep the pane closed and mark all news as already closed if it's the first time opening PixiEditor

PR #582, opened 2023-11-28
Autosaving and upgraded to .NET 8

### LINKED ISSUE #581: "Open from clipboard" causes a crash
1. Use lightshot to select something on the screen, ctrl+c to copy
2. Open pixieditor
3. Ctrl+K, open from clipboard
4. Pixieditor crashes

probably not lightshot specific, but don't wanna investigate rn
[crash-2023-10-28_19-55-53_242.zip](https://github.com/PixiEditor/PixiEditor/files/13196350/crash-2023-10-28_19-55-53_242.zip)

CPKreu:
Added periodic autosaving and upgraded to .NET 8

TODO

- [x] Add periodic autosaving
- [x] Allow disabling autosaving for the current document session
- [x] Add icons to the UI
- [x] Tighter integration with crash system
- [x] Last command before crash log
- [x] Fix #581


flabbet (changes requested):

flabbet:
Also I'm thinking, why save every x setting is a dropdown not number input?

flabbet:
Also seems like autosave timer isn't working?

src/PixiEditor/Models/DataHolders/CrashReport.cs
```diff
@@ -169,20 +199,22 @@ public void Save()
...
         int counter = 0;
-        Dictionary<string, string?> originalPaths = new();
+        Dictionary<string, CrashFilePathInfo> originalPaths = new();
         foreach (DocumentViewModel document in vm.DocumentManagerSubViewModel.Documents)
         {
             try
             {
                 string fileName = string.IsNullOrWhiteSpace(document.FullFilePath) ? "Unsaved" : Path.GetFileNameWithoutExtension(document.FullFilePath);
-                string nameInZip = $"{fileName}-{document.OpenedUTC}-{counter}.pixi".Replace(':', '_');
+                string nameInZip = $"{fileName}-{document.OpenedUTC.ToString(CultureInfo.InvariantCulture)}-{counter.ToString(CultureInfo.InvariantCulture)}.pixi"
```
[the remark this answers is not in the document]
flabbet: Nothing changed, why resolved?

flabbet (changes requested):
PixiEditor doesn't restore files that are manually saved after closing, while autosaved files are. I think this should work like VS Code, last session is restored no matter if it's unsaved by user or not.

flabbet (changes requested):
Autosave worked for me once and then stopped, no matter if I restarted the app or paused and resumed via AutosaveControl. Timer seems to be frozen

flabbet:
- Open PNG
- Draw something on it
- Close PixiEditor
- Reopen PixiEditor, recovered png doesn't have a name

Also unsaved documents lose thier translated name, so instead of "Bez Nazwy" (pl) on recover it becomes "Unnamed"

flabbet:
Weird thing happens with canvas:
1. Open attached .pixi file
2. Create a new file
3. Close PixiEditor
4. Reopen PixiEditor
5. Notice how not all layers are rendered unless you draw on the image
[Mando.zip](https://github.com/PixiEditor/PixiEditor/files/13539592/Mando.zip)

src/PixiEditor/NotifyableObject.cs
```diff
@@ -56,4 +65,24 @@ protected bool SetProperty<T>(ref T backingStore, T value, out T oldValue, Actio
...
+        {
+            return;
+        }
+
+        var exception = new CustomStackTraceException($"Do not call {methodName}() from a thread other than the UI thread. Calling for property '{property ?? "{ property name null }"}'. (Calling from '{currentThread.Name ?? "{ name null }"}' @{currentThread.ManagedThreadId})");
+        
+#if DEBUG
+        throw exception;
+#else
+        exception.GenerateStackTrace();
```
flabbet, line 84: Won't this cause any problems in runtime? Not crashing due to invalid thread seems like a dangerous thing

flabbet (approved):
I've gone ahead and fixed a few crashes I encountered while testing.
- First is kinda unrelated to PR, which is try-catching GetValue in preferences, because I ran autosave branch a while ago and since then one of the preferences structure changed, this caused PixiEditor to crash on start, I made it to return fallback value on exception
 - Second was related to non-ui thread set text during autosave, I added a few dispatchers to AutosaveViewModel

Also I removed a TextBlock from Autosave Control and made it a tooltip, because I feel like this text took too much space and was too distracting while it provided little value, an icon should be enough.

Generally, after testing it works as expected, but I'd like @Equbuxu to test it before merging as well.

flabbet:
Note to self: TODO .NET 8 installer support

src/PixiEditor/NotifyableObject.cs
CPKreu: Well, we shouldn't be able to implement wrong threads anymore as it will throw in debug

And sometimes another thread can update it without it crashing as far as I understand it so maybe let's not crash for something that doesn't need to be crashed

Equbuxu (changes requested):
```
I found an infinite loop
triggered by transform selection, and close without applying
seems like it infinitely autosaves 
and it also seems like Document.UpdateableChangeActive isn't set properly for transforming selection 

Equbuxu — 今日 01:24
and now I can't repro it
also for some reason files that were saved right before closing are marked as unsaved after reopening
while Untitled files that were never saved are marked as saved
but only sometimes
```

also passthrough action and modified/deleted on disk labels are still todo

Equbuxu (changes requested):
`CrashHelper.SendExceptionInfoToWebhook` must not use Task.Run (or anything with a separate thread), supposedly the reason for adding it was this:
> I remember it not sending the exception in certain scenerios cause it wasn't awaited
> weird, it shouldn't matter whether you await it or not. I can only imagine it happening if the app shuts down before it has a chance to run
> I think that was it

Also it seems like a good idea to go through the entire project and closely examite each `Task.Run`, they are dangerous

src/PixiEditor/Models/DocumentModels/ActionAccumulator.cs
```diff
@@ -39,6 +39,7 @@ public void AddFinishedActions(params IAction[] actions)
         queuedActions.AddRange(actions);
         queuedActions.Add(new ChangeBoundary_Action());
         TryExecuteAccumulatedActions();
+        document.AutosaveViewModel.HintFinishedAction();
```
Equbuxu, line 42: You can still pass `ChangeBoundary_Action` into AddActions manually, and it won't get picked up. Although it is a bit hard to account for

Equbuxu:
Looks like `DocumentViewModel.MarkAsSaved();` should be called using a passthough action, otherwise there is a problem of "you've added some actions, marked document as saved, then actions get processed, last change guid changes, document is no longer considered saved". This is the cause of * indicator appearing on images you've just opened

Equbuxu:
Strange inconsistency: 
1. open an existing .pixi, draw a rectangle and close pixieditor without applying transform -> Pixieditor closes without saving the rectangle
2. create a new .pixi, draw a rectangle, close without applying -> Pixieditor asks if you want to save the document
3. and ofc if you do apply the change, pixieditor will just save properly without asking in both cases
what's described in 1 also happens to never-saved-by-user files restored from previous session

I have also noticed pixieditor hanging and getting into an infinite autosaving loop during 2, but not sure if it's specific to 2 (probably not) or just something that can randomly happen. Ironically it happens inside "SafeAutosave". Might be caused by the -1 from AutosavePeriodMinutes ending up in the timer, with that timer then being enabled in `RestartTimers` and immediately firing (?) when an autosave happens during close

Equbuxu:
image
I just spent like 5 minutes trying to remember what the last checkbox does (wdym autosave if I already said I want to backup every 5 minutes) so apparently this wording sucks. I think all three features should have a label that's a lot more descriptive

Equbuxu:
Seems like autosave files never get deleted? I made some new documents, closed pixieditor, reopened it, then closed all documents without saving, closed pixieditor, and now I have zombie .pixi files laying in Temp\PixiEditor\Autosave

Equbuxu:
set your settings to this:
image
create a new document and wait for it to get autosaved. Then close pixieditor. You don't get prompted about having unsaved stuff, but nothing is restored after you reopen pixieditor (no shit, restore session thingie is disabled). So effectively you'be just lost your document, unless you know you can look for it in the temp/pixieditor/autosave directory

Equbuxu:
- With backup timer running, quickly create a few documents and close them right away. Wait for the interval of the timer. Once it elapses, the closed documents write autosaves into autosave directory. They must be sticking around in memory because of avalondock
- I want a seconds counter
- Would be nice to handle the case where pixieditor is killed (without it crashing), backups aren't restored in this case even though nothing is stopping us from restoring them

src/PixiEditor/Models/DocumentModels/ActionAccumulator.cs
Equbuxu: I decided that this doesn't really matter, saving during an updateable change shouldn't cause any issues apart from a potential lag spike, and this is fine for preventing the lag spike

flabbet:
Honestly at this point, this branch can be at most used as a reference

Issue #584, opened 2023-12-05
Deadlock when opening non-instance window with document open


CPKreu:
To repro

* Create new document
* Try opening a window which doesn't use instances like the preference window

PR #585, opened 2023-12-09
Hopefully fix MouseUpdateController for good

Equbuxu:
- #584

### DIFF
--- a/src/PixiEditor/Models/Controllers/MouseUpdateController.cs
+++ b/src/PixiEditor/Models/Controllers/MouseUpdateController.cs
@@ -16,6 +16,7 @@ public class MouseUpdateController : IDisposable
 
     private readonly FrameworkElement element;
     private readonly MouseEventHandler mouseMoveHandler;
+    
 
     public MouseUpdateController(FrameworkElement uiElement, MouseEventHandler onMouseMove)
     {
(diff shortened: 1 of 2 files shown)


flabbet (approved):
Deadlock is gone, seems to be working. However please test if amount of mouse events isn't bigger than previously. My laptop is kinda bad and performance is not the best, however it also happens on released PixiEditor versions, so I can't conclude if here is worse or not.

flabbet (changes requested):
Actually, nvm, there is one issue. Probably mouse thread isn't dying after closing.

To reproduce:
1. Run debug in Rider
2. close PixiEditor
3. Notice how Rider still shows that PixiEditor is running (or just check task manager under Rider processes)

Running other PixiEditor instance causes it to crash

src/PixiEditor/Models/Controllers/MouseUpdateController.cs
```diff
@@ -1,49 +1,126 @@
...
         {
-            element.MouseMove += OnMouseMove;
-        });
+            long lastThreadIter = Stopwatch.GetTimestamp();
+            
+            // abort if a new thread was created
+            while (!IsThreadShouldStop())
+            {
+                // call waitOne periodically instead of waiting infinitely to make sure we crash or exit when resetEvent is disposed
+                if (!resetEvent.WaitOne(300))
```
flabbet, line 66: You could make 300 a constant with a name, since I kinda don't know why its' 300

flabbet (approved):
Nice, seems to do the job well

PR #588, opened 2023-12-13
Add support to read DeluxePaint palettes

warrengalyen:


flabbet:
Could you please provide a few files to test the parser with?

warrengalyen:
Here you go. 
[DeluxePaint palettes.zip](https://github.com/PixiEditor/PixiEditor/files/13676346/DeluxePaint.palettes.zip)

Equbuxu (approved):
Looks good to me, works with your files as well as with files I exported from DeluxePaint myself (minus the extension in capital letters issue). Just curious, do you know of any modern software that uses these palette formats? I couldn't find any, so I resorted to running DeluxePaint in DOSBox

warrengalyen:
No, I only know of software that supports reading them, but the format itself was only used by DeluxePaint on Amiga and DOS. I exported the palettes directly as well directly from DeluxePaint. Just thought it would be nice to support since it was a pretty popular editor for pixel art back in the day.

flabbet (approved):

PR #594, opened 2023-12-15
Add support for CorelDraw 3.0 palettes

warrengalyen:
Test palettes:

[CorelDraw Palettes.zip](https://github.com/PixiEditor/PixiEditor/files/13679972/CorelDraw.Palettes.zip)

This PR also includes refactoring of palette import to support multiple formats of same extension, as it was a blocker.

### DIFF
--- a/src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
+++ b/src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
@@ -70,6 +70,8 @@ internal class CorelDrawPalParser : PaletteFileParser
         return new PaletteFileData(name, colors.ToArray());
     }
 
+    public override bool CanSave => false;
+
     // NOTE: saving data works but PixiEditor defaults to first .pal parser which
     //       is JASC so saving to an actual file is currently not available.
     //       Palette exporter needs to be refactored to support multiple formats of
--- a/src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
+++ b/src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
@@ -1,0 +1,218 @@
+﻿using System.IO;
+using System.Text;
+using PixiEditor.Extensions.Palettes;
+using PixiEditor.Extensions.Palettes.Parsers;
+
+namespace PixiEditor.Models.IO.PaletteParsers;
+
+internal class CorelDrawPalParser : PaletteFileParser
+{
+    public override string FileName { get; } = "CorelDRAW! 3.0 Palette";
+    public override string[] SupportedFileExtensions { get; } = { ".pal" };
+
+    // Default name to use for colors (color name is required by format)
+    private const string SWATCH_NAME = "PixiEditor Color";
+
+    public override async Task<PaletteFileData> Parse(string path)
+    {
+        try
+        {
+            return await ParseFile(path);
+        }
+        catch
+        {
+            return PaletteFileData.Corrupted;
+        }
+    }
+
+    private async Task<PaletteFileData> ParseFile(string path)
+    {
+        string name = Path.GetFileNameWithoutExtension(path);
+
+        List<PaletteColor> colors = new();
+
+        await using (Stream stream = File.OpenRead(path))
+        {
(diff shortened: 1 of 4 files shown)


src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
```diff
@@ -0,0 +1,220 @@
...
+
+    private async Task<PaletteFileData> ParseFile(string path)
+    {
+        string name = Path.GetFileNameWithoutExtension(path);
+
+        List<PaletteColor> colors = new();
+
+        await using (Stream stream = File.OpenRead(path))
+        {
+            using (TextReader reader = new StreamReader(stream, Encoding.ASCII))
```
flabbet, line 36: TextReader.ReadLine has async overload

src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        return new PaletteFileData(name, colors.ToArray());
+    }
+
+    public override bool CanSave => false;
+
+    // NOTE: saving data works but PixiEditor defaults to first .pal parser which
+    //       is JASC so saving to an actual file is currently not available.
+    //       Palette exporter needs to be refactored to support multiple formats of
+    //       same file extension.
```
flabbet, line 78: Due to this, I can't really merge your PR. Some of core contributors would be forced to actually implement this before next release, so stuff doesn't break. If you don't implement this yourself, then this PR will be on hold for unknown amount of time (until we add this feature)

src/PixiEditor/Views/Dialogs/PalettesBrowser.xaml.cs
```diff
@@ -652,14 +647,23 @@ private async void ImportFromFile_OnClick(object sender, RoutedEventArgs e)
...
         {
-            var data = await parser.Parse(fileName);
+            int index = 0;
+            foreach (var parser in parserList)
+            {
+                var data = await parser.Parse(fileName);
+                index++;
+
+                if (data.IsCorrupted && index == parserList.Count) return; // fail if none of the parsers in our list can read the file
+                if (data.IsCorrupted) continue; // skip to next parser if unable to read
```
flabbet, line 662: This might be enough for most cases, however I'd prioritize magic bytes check, if this fails (because some format might not have magic bytes), then I'd fallback to try parsing

src/PixiEditor/Views/UserControls/Palettes/PaletteViewer.xaml.cs
```diff
@@ -111,12 +109,24 @@ private async void ImportPalette_OnClick(object sender, RoutedEventArgs e)
...
+        if (parserList != null)
+        {
+            int index = 0;
+            foreach (var parser in parserList)
+            {
+                var data = await parser.Parse(fileName);
+                index++;
+
+                if ((data.IsCorrupted || data.Colors.Length == 0) && index == parserList.Count) return; // fail if none of the parsers in our list can read the file
+                if (data.IsCorrupted) continue; // skip to next parser if unable to read
```
flabbet, line 124: Same story as my previous comment, btw this could be extracted into one method, so the code isn't duplicated.

flabbet (changes requested):

src/PixiEditor/Models/IO/PaletteParsers/CorelDrawPalParser.cs
warrengalyen: I figured, that is why I set the save disabled attribute. The write code was let in place for future use.

src/PixiEditor/Views/Dialogs/PalettesBrowser.xaml.cs
warrengalyen: I'm just relying on the the parser to fail to read the file, which should be checking for magic bytes. I didn't want to change any of the other existing core code. It worked as I tested on several different .pal formats.

flabbet (changes requested):
I pushed a commit to this PR with support for saving palette when multiple parsers has the same extensions. Please add saving support for CorelDraw 3.0 palettes and I'll merge this.

flabbet (approved):
Everything seems good

PR #600, opened 2023-12-19
Hopefully now fix mouse controller for good

Equbuxu:


flabbet (changes requested):
Thread is not dying after PixiEditor is closed

flabbet (approved):

PR #607, opened 2024-01-12
Fix Jasc PSP palette parser

warrengalyen:
Since at least v20, Paint Shop Pro palette extensions default to ".PspPalette". For compatibility sake, I've add this extension to the parser so these files can be imported as well. Corel still uses the same old data format as JASC for these files so no other changes need be made. I've tested with the current version 2023. Export will still default to the .pal extension.


flabbet (approved):
Sorry for late response, thanks for the PR!

PR #613, opened 2024-03-16
Selecting the closest layer when deleting

Kuvrot:
### Selecting the closest layer when deleting (similar to GIMP)

https://github.com/PixiEditor/PixiEditor/assets/23508114/255f9649-a95b-4e3e-8717-51b0b7fadb86

I noticed that when deleting a layer, there was no layer selected after, and this make working uncomfortable because you couldn't  draw after deleting a layer, and you had to manually select the layer you wanted to work with.

The logic used was: 
* Search in the list of layers in the project.
* Locate the layer that is being deleted.
* Select the next layer above or below depending on the position of the layer that is being deleted.

### DIFF
--- a/src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
+++ b/src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
@@ -84,9 +84,7 @@ internal class LayersViewModel : SubViewModel<ViewModelMain>
 
                         break;
                 }
-
-            }
-
+             }
             doc.Operations.AddSoftSelectedMember(baseLayer.GuidValue);
             doc.InternalSetSelectedMember(baseLayer);
         }
--- a/src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
+++ b/src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Immutable;
+using System.Diagnostics;
+using System.Dynamic;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
@@ -15,6 +17,7 @@ using PixiEditor.Models.Dialogs;
 using PixiEditor.Models.Enums;
 using PixiEditor.Models.IO;
 using PixiEditor.Models.Localization;
+using PixiEditor.Parser;
 using PixiEditor.ViewModels.SubViewModels.Document;
 using PixiEditor.Views.UserControls.Layers;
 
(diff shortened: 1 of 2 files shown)


flabbet (changes requested):

src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -48,10 +50,44 @@ public bool CanDeleteSelected()
...
+                        baseLayer = doc.StructureHelper.GetAllLayers()[i];
+                    }
+                    else if (i == 0)
+                    {
+                        baseLayer = doc.StructureHelper.GetAllLayers().ToArray()[i+1];
+                    }
+
+                        break;
+                }
+             }
```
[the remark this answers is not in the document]
Kuvrot: Like this?

flabbet (changes requested):
Yes, however your solution doesn't work if you delete layers that are in folders, or generally delete a folder

Kuvrot:
I had to modify the `DocumentStructureModuleFile` in order to create a new Method called `GetAllMembers` that returned a list with all the members including folders and layers. Unfortunately I had to use another for loop inside for managing layers inside folders.

https://github.com/PixiEditor/PixiEditor/assets/23508114/90e125c6-ada2-425a-9da4-6fa818d8de04

src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -48,10 +53,80 @@ public bool CanDeleteSelected()
...
+                                    break; 
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            doc.InternalSetSelectedMember(baseLayer);
+            doc.Operations.AddSoftSelectedMember(baseLayer.GuidValue);
```
Equbuxu, line 128: "Soft" selection is used in the case where you have multiple layers selected. To handle drawing in that case, we need to choose a single layer to draw on. Because of that there are two types of selection, "Hard" selection (only one layer can be hard-selected) and "Soft" selection (any selected layers that aren't the main one will be "soft" selected).

It looks like you are trying to hard-select `baseLayer`, and immediately afterwards also soft-select it, which doesn't make sense (and there is a check in `DocumentUpdater.ProcessAddSoftSelectedMember` which ensures that you can't soft-select a hard-selected layer)

src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -48,10 +53,80 @@ public bool CanDeleteSelected()
...
     {
-        var member = Owner.DocumentManagerSubViewModel.ActiveDocument?.SelectedStructureMember;
+        var doc = Owner.DocumentManagerSubViewModel?.ActiveDocument;
+
+        var member = doc.SelectedStructureMember;
         if (member is null)
             return;
+
         member.Document.Operations.DeleteStructureMember(member.GuidValue);
+        var allLayers = doc.StructureHelper.GetAllMembers().ToArray();
```
Equbuxu, line 63: you can iterate over `doc.StructureRoot.Children` directly, no need to copy the structure into a separate list. You can get rid of both `GetAllMembers` overloads (one of them seems to be unused already?)

Edit: although, you likely won't need `doc.StructureRoot.Children` at all, see further comments

src/PixiEditor/ViewModels/SubViewModels/Main/LayersViewModel.cs
```diff
@@ -48,10 +53,80 @@ public bool CanDeleteSelected()
...
+                    else if (i == 0)
+                    {
+                        baseLayer = allLayers[i + 1];
+                    }
+
+                    break;
+                }
+                else
+                {
+                    //Managing elements inside folders
```
Equbuxu, line 94: Unless I'm missing something here (I haven't run the code, sry) this doesn't handle nested folders (since folders can contain other folders that can contain other folders..). I think you need an algorithm that goes something like:
1. Find the parent of the member being deleted (`DocumentStructureModule.FindChildAndParentOrThrow`)
2. If the parent has only one child (which is getting deleted) then select the parent
3. Otherwise, use a loop to find the closest child to the one getting deleted (like you are already doing).
4. Select found child

Equbuxu (changes requested):

flabbet:
Closing for inactivity

PR #621, opened 2024-07-20
move "if" from inside an "else" to merge into an "else if"

MichaelHinrichs:
See title. A minor refactoring.


flabbet:
unnecessary change

PR #622, opened 2024-07-20
Use a switch

MichaelHinrichs:
Replace ZoomMode "else if"s with a switch for faster code.


flabbet:
C# compiler is perfectly capable of this kind of optimizations under the hood. Changing if's to switches doesn't make any real difference in that case

PR #624, opened 2024-07-20
Throw if in throw if disposed

MichaelHinrichs:
Just doing what an issue on CodeFactor said to do.
"Use 'ObjectDisposedException.ThrowIf' instead of explicitly throwing a new exception instance."


flabbet:
This PR doesn't seem to provide any real value in the codebase

PR #625, opened 2024-07-20
Cleanup

MichaelHinrichs:
## Pull Requests
Clean up unnecessary whitespace and braces in DocumentViewModel.cs, PalettesBrowser.xaml.cs, LayersViewModel.cs, and CommandController.cs.


flabbet:
This PR doesn't provide any value, in fact in the codebase we wrap single line if bodies into brackets. Closing

PR #633, opened 2024-08-29
Node view models

CPKreu:
Added node view model to get rid of frontend stuff in backend and added dynamic inputs/name to the math node

Also fixed ellipse not using position for rasterization


src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/Shapes/Data/EllipseData.cs
```diff
@@ -16,7 +16,14 @@ public EllipseData(VecD center, VecD radius)
     
     public override void Rasterize(DrawingSurface drawingSurface)
     {
-        using ChunkyImage img = new ChunkyImage(new VecI((int)Radius.X * 2, (int)Radius.Y * 2));
+        var imageSize = new VecI((int)Radius.X * 2, (int)Radius.Y * 2) + (VecI)Center;
```
flabbet, line 19: I don't get this, why Image size should have Center component added? Size of the circle is a radius*2

src/PixiEditor/Models/DocumentModels/DocumentUpdater.cs
```diff
@@ -511,28 +505,43 @@ private void ClearSelectedKeyFrames(ClearSelectedKeyFrames_PassthroughAction inf
...
-    private void ProcessCreateNode<T>(CreateNode_ChangeInfo info) where T : NodeViewModel, new()
+    private void ProcessCreateNode(CreateNode_ChangeInfo info)
     {
-        T node = new T()
-        {
-            InternalName = info.InternalName, Id = info.Id, Document = (DocumentViewModel)doc, Internals = helper
-        };
+        var nodeType = info.Metadata.NodeType;
+        
+        var ns = nodeType.Namespace.Replace("PixiEditor.ChangeableDocument.Changeables.Graph.Nodes", "PixiEditor.ViewModels.Document.Nodes");
```
flabbet, line 512: I hate this being hardcoded, maybe query last dot and substring the type name, then append "PixiEditor.Viewmodels.Document.Nodes" there?

src/PixiEditor/Models/Handlers/IDocument.cs
```diff
@@ -31,8 +31,8 @@ internal interface IDocument : IHandler
     public bool AllChangesSaved { get; }
     public string CoordinatesString { get; set; }
     public IReadOnlyCollection<IStructureMemberHandler> SoftSelectedStructureMembers { get; }
-    public ILayerHandlerFactory LayerHandlerFactory { get; }
-    public IFolderHandlerFactory FolderHandlerFactory { get; }
+    //public ILayerHandlerFactory LayerHandlerFactory { get; }
+    //public IFolderHandlerFactory FolderHandlerFactory { get; }
```
flabbet, line 35: You can remove it if it's not needed or add a comment why it is commented out

src/PixiEditor/ViewModels/Document/DocumentViewModel.cs
```diff
@@ -209,8 +210,8 @@ public Texture PreviewSurface
     IDocumentOperations IDocument.Operations => Operations;
     ITransformHandler IDocument.TransformHandler => TransformViewModel;
     ILineOverlayHandler IDocument.LineToolOverlayHandler => LineToolOverlayViewModel;
-    public ILayerHandlerFactory LayerHandlerFactory { get; }
-    public IFolderHandlerFactory FolderHandlerFactory { get; }
+    //public ILayerHandlerFactory LayerHandlerFactory { get; }
+    //public IFolderHandlerFactory FolderHandlerFactory { get; }
```
flabbet, line 214: Same story as previous comment

src/PixiEditor/ViewModels/Document/DocumentViewModel.cs
```diff
@@ -224,8 +225,8 @@ private DocumentViewModel()
         EventInlet = new DocumentEventsModule(this, Internals);
         Operations = new DocumentOperationsModule(this, Internals);
 
-        LayerHandlerFactory = new LayerHandlerFactory(this);
-        FolderHandlerFactory = new FolderHandlerFactory(this);
+        //LayerHandlerFactory = new LayerHandlerFactory(this);
+        //FolderHandlerFactory = new FolderHandlerFactory(this);
```
flabbet, line 229: .

src/PixiEditor/ViewModels/Document/DocumentViewModel.cs
```diff
@@ -537,7 +538,7 @@ public OneOf<Error, Surface> TryRenderWholeImage(KeyFrameTime frameTime)
         IStructureMemberHandler? layerToExtractFrom = null)
     {
         layerToExtractFrom ??= SelectedStructureMember;
-        if (layerToExtractFrom is not LayerViewModel layerVm)
+        if (layerToExtractFrom is not ImageLayerNodeViewModel layerVm)
```
flabbet, line 541: The difference between this change and the previous is that `LayerViewModel` was a generic "Layer" view model, while you added specific layer type `ImageLayerNode`, It doesn't matter at this moment, but will when we have more layer types.

src/PixiEditor/ViewModels/Document/DocumentViewModel.cs
```diff
@@ -667,7 +668,7 @@ public Color PickColorFromCanvas(VecI pos, DocumentScope scope, KeyFrameTime fra
                         _ => Colors.Transparent);
             }
 
-            if (SelectedStructureMember is not LayerViewModel layerVm)
+            if (SelectedStructureMember is not ImageLayerNodeViewModel layerVm)
```
flabbet, line 671: Same as above

src/PixiEditor/ViewModels/Document/FolderHandlerFactory.cs
```diff
@@ -1,21 +1,22 @@
...
-    {
-        Document = document;
-    }
-
-    public IFolderHandler CreateFolderHandler(DocumentInternalParts helper, Guid infoGuidValue)
-    {
-        return new FolderViewModel(Document, helper, infoGuidValue);
-    }
-}
+﻿// using PixiEditor.Helpers;
```
flabbet, line 1: Git history have this in case we need it, delete or comment why it is commented

src/PixiEditor/ViewModels/Document/LayerHandlerFactory.cs
```diff
@@ -1,21 +1,22 @@
...
-    {
-        Document = document;
-    }
-
-    public ILayerHandler CreateLayerHandler(DocumentInternalParts helper, Guid infoGuidValue)
-    {
-        return new LayerViewModel(Document, helper, infoGuidValue);
-    }
-}
+﻿// using PixiEditor.Helpers;
```
flabbet, line 1: same

src/PixiEditor/ViewModels/Document/Nodes/Animable/TimeNodeViewModel.cs
```diff
@@ -0,0 +1,7 @@
+using PixiEditor.ChangeableDocument.Changeables.Graph.Nodes.Animable;
+using PixiEditor.ViewModels.Nodes;
+
+namespace PixiEditor.ViewModels.Document.Nodes.Animable;
+
+[NodeViewModel("TIME_NODE", "ANIMATION", "\uE900")]
```
flabbet, line 6: It's my personal preference, feel free to ignore it if you disagree, but `NodeViewModelAttribute` name could be `NodeSettings` or something, we already have `NodeViewModel<>` below, so we know it's a view model and settings feels more fitting.

src/PixiEditor/ViewModels/Nodes/NodeTypeInfo.cs
```diff
@@ -20,26 +18,42 @@ public class NodeTypeInfo
...
         DisplayName = attribute.DisplayName;
         PickerName = attribute.PickerName;
         Category = attribute.Category ?? "";
+        Icon = attribute.Icon;
+
+        FinalPickerName = (PickerName ?? DisplayName);
+    }
 
-        if (NodeIcons.IconMap.TryGetValue(type, out var icon))
+    private Type GetNodeType(Type? baseType)
```
flabbet, line 42: This method looks _extremely_ unsafe. Argument is nullable, yet no null check is inside. Also this method will freeze if base type is not assignable to Node, sure this scenario shouldn't happen. But if you do `while (baseType != null)` it should be enough.

src/PixiEditor/ViewModels/SubViewModels/LayersViewModel.cs
```diff
@@ -131,7 +132,7 @@ public bool CanCreateNewMember()
     public void ToggleLockTransparency()
     {
         var member = Owner.DocumentManagerSubViewModel.ActiveDocument?.SelectedStructureMember;
-        if (member is not LayerViewModel layerVm)
+        if (member is not ImageLayerNodeViewModel layerVm)
```
flabbet, line 135: I think checking against `ImageLayerNodeViewModel` is justified here, since I don't think other layer types would use lock transparency. Gj

src/PixiEditor/ViewModels/SubViewModels/LayersViewModel.cs
```diff
@@ -171,7 +172,7 @@ public void OpacitySliderSet(double value)
     public void DuplicateLayer()
     {
         var member = Owner.DocumentManagerSubViewModel.ActiveDocument?.SelectedStructureMember;
-        if (member is not LayerViewModel layerVM)
+        if (member is not ImageLayerNodeViewModel layerVM)
```
flabbet, line 175: Checking against generic LayerViewModel would be better here, ability to duplicate any layer is desired

src/PixiEditor/ViewModels/SubViewModels/LayersViewModel.cs
```diff
@@ -180,7 +181,7 @@ public void DuplicateLayer()
     public bool SelectedMemberIsLayer(object property)
     {
         var member = Owner.DocumentManagerSubViewModel.ActiveDocument?.SelectedStructureMember;
-        return member is LayerViewModel;
+        return member is ImageLayerNodeViewModel;
```
flabbet, line 184: `member is LayerNodeViewModel`

src/PixiEditor/Views/Layers/LayerControl.axaml.cs
```diff
@@ -7,17 +7,18 @@
...
 {
     public static string? LayerControlDataName = typeof(LayerControl).FullName;
 
-    public static readonly StyledProperty<LayerViewModel> LayerProperty =
-        AvaloniaProperty.Register<LayerControl, LayerViewModel>(nameof(Layer));
+    public static readonly StyledProperty<ImageLayerNodeViewModel> LayerProperty =
+        AvaloniaProperty.Register<LayerControl, ImageLayerNodeViewModel>(nameof(Layer));
 
-    public LayerViewModel Layer
+    public ImageLayerNodeViewModel Layer
```
flabbet, line 21: ImageLayerNodeViewModel at the moment is all right

flabbet (changes requested):
Overall great job! I didn't find any major issues. Just a few small ones

src/PixiEditor/ViewModels/Document/Nodes/Animable/TimeNodeViewModel.cs
CPKreu: I think C# devs did the alias thing (being able to use attributes without the attribute append) just so you could give a class a attribute with the same name.
I think the name fits just as well as it is an info attribute about the view model

flabbet (approved):

PR #634, opened 2024-08-29
Hsv/Hsl Support for Combine/Separate Color nodes

CPKreu:
Added support for Hsv and Hsl by adding built-in functions to the shader builder


src/PixiEditor.DrawingApi.Core/Shaders/Generation/BuiltInFunctions.cs
```diff
@@ -0,0 +1,161 @@
...
+        
+        return builder.ToString();
+
+        void AppendIf(IBuiltInFunction function)
+        {
+            if (usedFunctions.Contains(function))
+            {
+                builder.AppendLine(function.FullSource);
+            }
+        }
```
flabbet, line 46: Can't this just be 

```cs
foreach(var func in usedFunctions)
{
    AppendIf(func); // or just Append, since usedFunctions is probably already distinct list
}
```

src/PixiEditor.ChangeableDocument/Changeables/Graph/Context/FuncContext.cs
```diff
@@ -98,17 +99,58 @@ public Half4 NewHalf4(Expression r, Expression g, Expression b, Expression a)
...
+            constantHalf4.ConstantValue = Color.FromHsv((float)hValue, (float)sValue, (float)vValue, aByte);
+            return constantHalf4;
+        }
+
+        if (Half4Float1Accessor.AllAccessSame(h, s, v, a, out var half4))
+        {
+            return Builder.AssignNewHalf4(Builder.Functions.GetHsvToRgb(half4));
+        }
+        
+        return Builder.AssignNewHalf4(Builder.Functions.GetHsvToRgb(h, s, v, a));
```
flabbet, line 131: So if `h`, `s`, `v`, `a` components are part of the same half4, function produces new half4 based on the parent half4 instead of child components? What advantage does it have over just passing the components itself (second return)?

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CombineSeparate/SeparateColorNode.cs
```diff
@@ -19,39 +22,61 @@ public class SeparateColorNode : Node
...
+    private Half4 GetHsva(FuncContext ctx)
     {
-        Half4 target = null;
-        if (lastContext == ctx)
+        if (!ctx.HasContext && ctx.GetValue(Color) is Half4 constantColor)
         {
-            target = lastColor;
+            var variable = new Half4(string.Empty);
+            constantColor.ConstantValue.ToHsv(out float h, out float s, out float l);
+            variable.ConstantValue = new Color((byte)(h * 255), (byte)(s * 255), (byte)(l * 255), constantColor.ConstantValue.A);
```
flabbet, line 56: So the variable is Hsva color value, but constant values are multiplied like it's rgb?

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CombineSeparate/SeparateColorNode.cs
```diff
@@ -19,39 +22,61 @@ public class SeparateColorNode : Node
...
+    {
+        if (!ctx.HasContext && ctx.GetValue(Color) is Half4 constantColor)
         {
-            target = Color.Value(ctx);
+            var variable = new Half4(string.Empty);
+            constantColor.ConstantValue.ToHsl(out float h, out float s, out float l);
+            variable.ConstantValue = new Color((byte)(h * 255), (byte)(s * 255), (byte)(l * 255), constantColor.ConstantValue.A);
+            
+            return variable;
         }
```
flabbet, line 75: I kind of don't like this approach, Node shouldn't be concerned about dealing with shader builder. Originally my idea was that node writes C# "pseudocode" that gets then converted into shader, you shouldn't know that underneath you are dealing with constant values. This is a previous implementation of this node

```cs 
R = CreateFuncOutput<Float1>(nameof(R), "R", ctx => GetColor(ctx).R);
        G = CreateFuncOutput<Float1>(nameof(G), "G", ctx => GetColor(ctx).G);
        B = CreateFuncOutput<Float1>(nameof(B), "B", ctx => GetColor(ctx).B);
        A = CreateFuncOutput<Float1>(nameof(A), "A", ctx => GetColor(ctx).A);
...
 private Half4 GetColor(FuncContext ctx)
    {
        Half4 target = null;
        if (lastContext == ctx)
        {
            target = lastColor;
        }
        else
        {
            target = (Half4)ctx.GetValue(Color); 
        }
        
        lastColor = target;
        lastContext = ctx;
        return lastColor;
    }
```

With your changes I'd expect to do something like 

```cs
 private Half4 GetColor(FuncContext ctx) =>
        Mode.Value switch
        {
            CombineSeparateColorMode.RGB => GetRgba(ctx),
            CombineSeparateColorMode.HSV => GetHsva(ctx),
            CombineSeparateColorMode.HSL => GetHsla(ctx)
        };

    private Half4 GetRgba(FuncContext ctx)
    {
        return (Half4)ctx.GetValue(Color);
    }
    
    private Half4  GetHsva(FuncContext ctx)
    {
        var rgba = GetRgba(ctx);
        var hsva = ShaderMath.RgbaToHsva(rgba); // could be also ShaderColorMath or sth, just an example
        return ctx.NewHalf4(hsva);
    }
```

FuncContext is responsible for efficient "communication" between shader builder and node. So for example when you are getting value in node via FuncContext, it ensures that you are setting values via uniforms, not hardcoded values etc. You should use it in node instead of shader builder

On the other hand Shader builder is "dumb" it shouldn't do any optimizations, it just follows your orders hard

flabbet: My best advice is to follow the convention that `MathNode` does

flabbet (changes requested):
Almost all beta examples are broken

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CombineSeparate/SeparateColorNode.cs
CPKreu: if it wasn't multiplied you would always get a output of either O or 1 per component

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CombineSeparate/SeparateColorNode.cs
CPKreu: I don't remember why I didn't put it in FuncContext, I remember there being a issue
imma check when I get to it

flabbet (changes requested):
One more thing

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CombineSeparate/CombineColorNode.cs
```diff
@@ -26,22 +35,49 @@ public CombineColorNode()
         Color = CreateFuncOutput(nameof(Color), "COLOR", GetColor);
         Mode = CreateInput("Mode", "MODE", CombineSeparateColorMode.RGB);
 
-        R = CreateFuncInput<Float1>("R", "R", 0d);
-        G = CreateFuncInput<Float1>("G", "G", 0d);
-        B = CreateFuncInput<Float1>("B", "B", 0d);
+        V1 = CreateFuncInput<Float1>("R", "R", 0d);
```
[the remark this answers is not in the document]
CPKreu: I had it planned, shouldn't be too hard but I had to wait for view model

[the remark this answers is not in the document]
CPKreu: done

src/PixiEditor.DrawingApi.Core/Shaders/Generation/Expressions/Float1.cs
```diff
@@ -12,8 +12,60 @@ public class Float1(string name) : ShaderExpressionVariable<double>(name)
...
 
-    public static implicit operator Float1(double value) => new Float1("") { ConstantValue = value };
+    public static implicit operator Float1(double value) => new("") { ConstantValue = value };
 
     public static explicit operator double(Float1 value) => value.ConstantValue;
+
+    public byte FullSizeByteConstant => (byte)(ConstantValue * 255);
+}
+
+public class Half4Float1Accessor : Float1
```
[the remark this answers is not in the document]
CPKreu: done

src/PixiEditor.DrawingApi.Core/Shaders/Generation/Expressions/Float1.cs
```diff
@@ -12,8 +12,60 @@ public class Float1(string name) : ShaderExpressionVariable<double>(name)
 
     public override Expression? OverrideExpression { get; set; }
 
-    public static implicit operator Float1(double value) => new Float1("") { ConstantValue = value };
+    public static implicit operator Float1(double value) => new("") { ConstantValue = value };
 
     public static explicit operator double(Float1 value) => value.ConstantValue;
+
+    public byte FullSizeByteConstant => (byte)(ConstantValue * 255);
```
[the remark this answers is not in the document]
CPKreu: done

flabbet (approved):

PR #635, opened 2024-09-01
Send OS to analytics and recover analytics session after crash

CPKreu:
Send a Platform ID (e.g. "Windows", "Linux") and a full name "Microsoft Windows NT 10.0.19044.0" to the analytics API

And analytics sessions are now recovered after a crash

### DIFF
--- a/src/PixiEditor.Linux/LinuxOperatingSystem.cs
+++ b/src/PixiEditor.Linux/LinuxOperatingSystem.cs
@@ -1,4 +1,5 @@
 ﻿using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Input;
 using Avalonia.Threading;
 using PixiEditor.OperatingSystem;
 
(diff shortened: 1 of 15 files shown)


src/PixiEditor.Linux/LinuxOperatingSystem.cs
```diff
@@ -24,4 +27,43 @@ public bool HandleNewInstance(Dispatcher? dispatcher, Action<string> openInExist
...
+            if (!File.Exists(FilePath))
+            {
+                return new LinuxOSInformation(null, null, false);
+            }
+            
+            var lines = File.ReadAllLines(FilePath).Select<string, (string Key, string Value)>(x =>
+            {
+                var separatorIndex = x.IndexOf('=');
+                return (x[..separatorIndex], x[(separatorIndex + 1)..]);
+            }).ToList();
```
flabbet, line 53: This is unreadable for me, I have no clue what value are you getting here

src/PixiEditor/Models/ExceptionHandling/CrashReport.cs
```diff
@@ -115,6 +116,8 @@ private static void AppendStateInfo(StringBuilder builder)
         builder
             .AppendLine("Environment:")
             .AppendLine($"  Thread Count: {GetFormatted(() => Process.GetCurrentProcess().Threads.Count)}")
+            .AppendLine("Analytics:")
```
flabbet, line 119: While we are at it, can you do anonymizing path in crash reports? It's better to make sure that everything is fully anonymous

flabbet (changes requested):

flabbet (approved):

PR #712, opened 2025-01-03
Only send periodic performance reports for a 1 hour span

CPKreu:
This will minimize PeriodicPerformanceReports to 80 events per session


flabbet (approved):
lgtm

PR #859, opened 2025-03-23
initial grid line size config work

wareya:

Add inputs to the viewport overlay that let the user resize the grid. Only added there, and not saved to project or settings.

The pixelsize and rows/columns properties were redundant with each other, so I replaced the rows/columns properties with grid size properties.

VS deleted the indentation on empty lines when I saved the files I modified, so the diff is slightly bloated. I don't know how to configure it to not do this. I personally prefer having indentation on empty lines.

Localization data for text label only added for English.

https://github.com/user-attachments/assets/8d1c271f-2428-4a3c-bbac-50b4d8387985

## SELECT BELOW TO CONTINUE


src/PixiEditor/Views/Main/ViewportControls/Viewport.axaml
```diff
@@ -133,6 +134,26 @@
...
+                        <StackPanel Spacing="10" HorizontalAlignment="Center" Orientation="Horizontal">
+                            <TextBlock VerticalAlignment="Center"
+                                       ui:Translator.Key="GRIDLINES_YSIZE" />
+                            <input:NumberInput Width="32"
+                                         HorizontalAlignment="Left"
+                                         Min="1" Decimals="0"
+                                         Max="1024"
+                                         Value="{Binding GridLinesYSize, RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=viewportControls:Viewport}, Mode=TwoWay}" />
+                        </StackPanel>
+                        <Separator />
```
flabbet, line 156: There shouldn't be a separator between Grid X size and Grid Y size in UI, I'd actually do it similarly to render preview option, so label on top and two inputs below like:
```
________________
    Grid Size: 
[    ]  [    ] 
-----------------
```
I believe there is no need to put X: before first input and Y: before second since it's pretty obvious, but it depends on how it will look.

Additionally, the inputs should be disabled if grid lines are disabled.

flabbet: I would also enable decimals since vectors are not confined to pixel grid

src/PixiEditor/Views/Overlays/GridLinesOverlay.cs
```diff
@@ -103,4 +105,11 @@ private static void OnIsVisibleChanged(AvaloniaPropertyChangedEventArgs<bool> e)
             gridLines.Refresh();
         }
     }
```
flabbet, line 107: As per to my previous comment, I think we should allow floating point grid sizes and this won't work with them.

flabbet (changes requested):
I didn't test it yet, just looked at the code. I'll do it tomorrow, but isn't disappearing grid lines a problem if you zoom it out enough? It was made like that because grid lines with 1x1 dimensions are too dense at a certain distance

wareya:
> I didn't test it yet, just looked at the code. I'll do it tomorrow, but isn't disappearing grid lines a problem if you zoom it out enough? It was made like that because grid lines with 1x1 dimensions are too dense at a certain distance

Good point, I'll take a look at the code that hides them when zooming out.

src/PixiEditor/Views/Main/ViewportControls/Viewport.axaml
wareya: image

How's this?

I'm not sure one way or the other about disabling the inputs when the grid is disabled. I personally tend to get annoyed when I can't configure the numbers inside of settings that are disabled.

Also, I've disabled decimals, but now it's non-obvious that the easiest way to get an integer number of pixels with the mouse is by using the scroll wheel (instead of clicking and dragging). Does the UI system have some way to make it use integers when using the mouse, but allow typing in decimals?

src/PixiEditor/Views/Overlays/GridLinesOverlay.cs
wareya: image
🤔 I made sure to use doubles instead of ints and it seems to work once I made the UI XML code not have `decimals="0"`. The looping `i` is an integer, but it gets coerced to double when multiplied by `columnWidth`/`rowHeight`.

src/PixiEditor/Views/Main/ViewportControls/Viewport.axaml
flabbet: Yeah, the UI looks better now, however now that I think about is, it might be too small for bigger numbers and will overflow? 

About disabling inputs; fair enough, if it's frustrating for you I guess there are more people who'd feel like that, you can leave it as it is.

There is no rounding/step option for mouse dragging, but feel free to add it.

src/PixiEditor/Views/Overlays/GridLinesOverlay.cs
flabbet: Ah yeah, it will work, the only thing I think might not is the last line due to truncation for some values, but I might be wrong

src/PixiEditor/Views/Main/ViewportControls/Viewport.axaml
wareya: image

How's this?

flabbet: I guess this solution is the most practical. You can also play around using `SizeInput` instead of `NumberInput` it has [px] after the text, it might fit better, but it's up to you. It's good enough now

src/PixiEditor/Views/Overlays/GridLinesOverlay.cs
```diff
@@ -56,6 +56,8 @@ public int Rows
     static GridLinesOverlay()
     {
         IsVisibleProperty.Changed.Subscribe(OnIsVisibleChanged);
+        GridXSizeProperty.Changed.Subscribe(OnNumberChanged);
+        GridYSizeProperty.Changed.Subscribe(OnNumberChanged);
```
flabbet, line 60: fyi you can also use 
```cs
 static GridLinesOverlay()
    {
        IsVisibleProperty.Changed.Subscribe(OnIsVisibleChanged);
        AffectsOverlayRender(GridXSizeProperty, GridYSizeProperty);
    }
```

flabbet (approved):

flabbet:
Thanks for you contribution!

PR #892, opened 2025-04-12
Fixed non latin crash report file paths

CPKreu:

The previous regex only captured A-z 0-9 characters when parsing program arguments which did not capture cyrillic characters

This fixes PixiAnalytics Issue 48: DirectoryNotFoundException

## SELECT BELOW TO CONTINUE


flabbet (approved):
LGTM, interestingly, this regex wouldn't work in JS or PHP, since `\w` is equivalent to `[a-zA-Z0-9_]` there

PR #961, opened 2025-05-29
Fix passing null to string in InvalidFileTypeException

CPKreu:

At the moment, OpenRegularImage throws a `PixiEditor.Exceptions.InvalidFileTypeException: Invalid file type ''` because `fileType` is always null after the null check

```
        var fileType = SupportedFilesHelper.ParseImageFormat(Path.GetExtension(path));

        if (fileType != null)
        {
            var fileSize = new FileInfo(path).Length;
            Analytics.SendOpenFile(fileType, fileSize, doc.SizeBindable);
        }
        else
        {
            CrashHelper.SendExceptionInfo(new InvalidFileTypeException(default,
                $"Invalid file type '{fileType}'"));
        }
```

The PR fixes it by passing the file extension from `Path.GetExtension(path)` instead

## SELECT BELOW TO CONTINUE

### DIFF
--- a/src/PixiEditor/ViewModels/SubViewModels/FileViewModel.cs
+++ b/src/PixiEditor/ViewModels/SubViewModels/FileViewModel.cs
@@ -383,7 +383,8 @@ internal class FileViewModel : SubViewModel<ViewModelMain>
 
         AddRecentlyOpened(path);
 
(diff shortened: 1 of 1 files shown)


flabbet (approved):
lgtm

PR #1060, opened 2025-08-09
Added French Language and Flag

BetoJava:
Not much to explain, here is the translation into French, along with its flag :)


flabbet:
Hey, thank you for your contribution! Unfortunately we already have active French translators. 

And we're actually handling all our localization trough POEditor. Join https://discord.gg/tzkQFDkqQS in case you're interested in translating in the future.

PR #1070, opened 2025-08-11
Add Voronoi Noise Type Into Noise Node

starman:
Added option to generate Voronoi noise with noise node.

voronoi1

voronoi2

voronoi3

## SELECT BELOW TO CONTINUE


starman:
Hello, I added some inputs that will allow for more customization within the Voronoi noise node:


Feature (specifies which distance metric is used for generating the pattern) 
 - F1 (Distance to the closest feature point)
 - F2 (Distance to the second-closest feature point)
 - F2-F1 (Difference between the second-closest and closest distance)


Randomness (controls how "randomized" the placement of feature points is within each cell)
 - 0.0 - centered feature points (regular grid)
 - 1.0 - standard Voronoi pattern

Angle Offset (adjusts the offset of the pseudo-random angle for each feature point)

https://github.com/user-attachments/assets/da4cbbdb-298d-4020-b7e4-f0e185bbcbab

flabbet (commented):
Overall, I like the effect. Good job!

I actually haven't noticed that, but in Noise node shaders are not properly disposed.

Can you add that? 

On Line 89 `paint?.Shader?.Dispose()`
and on line 127 the same. Basically everywhere after `SelectShader()` is called

starman:
I pushed the requested changes. Please let me know if I used `.WithUpdatedUniforms(uniforms)` properly.

PR #1150, opened 2025-09-01
Add posterization node

starman:

Added posterization effect node

### Node inputs:
Mode(RGB, Luminance)
Levels(2 - 256)
Use document color space(sRGB if unchecked)

https://github.com/user-attachments/assets/e6f0df05-8a2d-4619-94f1-d309ce1e23a0

gradient2

gradient2

## SELECT BELOW TO CONTINUE


starman:
Color Space input is now an enum.

color_space_type

src/PixiEditor/Data/Localization/Languages/en.json
```diff
@@ -848,6 +848,10 @@
   "TYPE": "Type",
   "EFFECTS": "Effects",
   "OUTLINE_NODE": "Outline",
+  "POSTERIZATION_NODE": "Posterization",
```
flabbet, line 851: Change to "Posterize"

src/PixiEditor/ViewModels/Document/Nodes/Effects/PosterizationNodeViewModel.cs
```diff
@@ -0,0 +1,11 @@
+﻿using PixiEditor.ChangeableDocument.Changeables.Graph.Nodes.Effects;
+using PixiEditor.ViewModels.Nodes;
+
+namespace PixiEditor.ViewModels.Document.Nodes.Effects;
+
+[NodeViewModel("POSTERIZATION_NODE", "EFFECTS", PixiPerfectIcons.ColorPalette)]
```
flabbet, line 6: ColorPalette icon is already used in `Color` node. We are using lucide icons https://lucide.dev/icons/

To add an icon, open icomoon.app, drag [defs.svg](https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor.UI.Common/Fonts/defs.svg) and add your selected icon (you may need to flatten it in some app like Figma)

Give it a name in export, download the font and replace ttf and svg, then run `gen-icons.py`

flabbet (changes requested):
Overall, the effect is good. Before I merge, I would like you to check if it is possible to make this node a image filter, instead of render node. So it behaves like Shadow Filter or Gaussian Blur filter node. If it's not possible then fine, but filter experience is better, it's worth checking out

starman:
I updated the display name and icon.

As for changing this node to an image filter, I wasn’t able to get it to work. It might be possible through another approach, but the only lead I found is Skia’s `SkImageFilter::RuntimeShader`, which as of now, doesn’t appear to be exposed in SkiaSharp.

flabbet (approved):

Issue #1203, opened 2025-09-26
[BUG]: Node not rendering correct when sockets number changed.


heroboy:
### Bug Description

<img width="529" height="332" alt="Image" src="https://github.com/user-attachments/assets/0f2138d6-b87e-4057-b5ea-36a62ce26ae7" />

### Steps how to trigger the issue

```bash
1. Create a Math node
2. Connect something to X and Y
3. Change Math Mode to sin
```

### How did you download the software?

Steam

### Version Channel

Stable

### Version Number

2.0.1.14

### Operating System

Windows

### OS Version

win11

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

flabbet:
Thank you for the report. I fixed the issue and it's already on [development channel 2.0.1.15-dev](https://pixieditor.net/docs/open-beta/)

PR #1207, opened 2025-09-29
Dont render connection if socket invisible

flabbet:

#1203 

## SELECT BELOW TO CONTINUE


CPKreu (approved):
Could use pattern matching to check null and false at the same time
But it's good as is

PR #1256, opened 2025-11-08
Magic Wand: Hold and Drag + Modifiers

### LINKED ISSUE #843: [FEATURE]: Hold and drag magic wand
### Feature Description

Holding left mouse button when using flood fill makes contiguous fills, this is not a case for magic wand. It should behave similarily.

Related file: https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor.ChangeableDocument/Changes/Selection/MagicWand/MagicWand_Change.cs

### Steps how to use this feature

```bash
Holding left mouse button and dragging magic wand should execute magic wand tool, it should happen like that for every mode.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Francespo:
 
The changes in this PR close #843 

Now, holding the LMB executes the Magic Wand tool according to the current selected mode.
Also, the behaviour of the Magic Wand tool can be modified by pressing the shift/ctrl modifier. The action label on the bottom left of the screen updates accordingly (a proper description is still missing, though)

https://github.com/user-attachments/assets/a4aa45fa-8cca-4dbb-ab16-02fa90662d69

## SELECT BELOW TO CONTINUE

EDIT: added 'close' keyword

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 <img width="50%" align="center" src="https://github.com/user-attachments/assets/bd08c8bd-f610-449d-b1e2-6a990e562518">
 
-## The only 2D Graphics Editor you'll ever need
+## The All-in-One Editor for 2D
 
 **PixiEditor** is a universal 2D editor that was made to provide you with tools and features for all your 2D needs. Create beautiful sprites for your games, animations, edit images, create logos. All packed up in an intuitive and familiar interface.
 


src/PixiEditor/ViewModels/Tools/Tools/MagicWandToolViewModel.cs
```diff
@@ -45,4 +46,23 @@ public override void UseTool(VecD pos)
...
+    public override void KeyChanged(bool ctrlIsDown, bool shiftIsDown, bool altIsDown, Key argsKey)
+    {
+        if (shiftIsDown)
+        {
+            ActionDisplay = new LocalizedString("MAGIC_WAND_ACTION_DISPLAY_SHIFT");
+            KeyModifierselectionMode = SelectionMode.Add;
+        }
+        else if (ctrlIsDown)
+        {
+            ActionDisplay = new LocalizedString("MAGIC_WAND_ACTION_DISPLAY_CTRL");
```
flabbet: Please add this key to en.json

src/PixiEditor/Models/DocumentModels/UpdateableChangeExecutors/MagicWandToolExecutor.cs
```diff
@@ -25,27 +25,45 @@ public override ExecutionState Start()
...
-        internals!.ActionAccumulator.AddActions(new MagicWand_Action(memberGuids, pos, mode, tolerance, document!.AnimationHandler.ActiveFrameBindable));
+        AddUpdateAction(pos);
 
         return ExecutionState.Success;
     }
 
+    public override void OnPixelPositionChange(VecI pos)
+    {
+        AddUpdateAction(pos);
+        internals!.ActionAccumulator.AddActions(new ChangeBoundary_Action());
```
flabbet: Do not use `ChangeBoundary_Action`, use `AddFinishedActions` instead. Although I don't think it should be finished here. One user drag should be considered one undo change in that case imo. 

You've correctly made MagicWand an `UpdateableChange`, so multiple changes are part of one change. If you change boundary on each movement, you remove the benefit of using `UpdateableChange`

flabbet (changes requested):

flabbet (changes requested):
Overall looks good, only one small change left

src/PixiEditor/Models/DocumentModels/UpdateableChangeExecutors/MagicWandToolExecutor.cs
flabbet: I am not sure why, I'd have to debug. Change_Boundary tells the undo system that undo packet should be closed. A packet can contain multiple changes and undo/redo them together. 

Some updateable changes also have logic that merges multiple changes into one.

Issue #1145, opened 2025-08-28
[BUG]: libSkiaSharp.so: undefined symbol: uuid_parse


kero990:
### Bug Description

I compiled the project according to the official compilation tutorial, but encountered errors during runtime: 

`/root/PixiEditor/src/PixiEditor.Desktop/bin/Debug/net8.0/linux-arm64/PixiEditor.Desktop: symbol lookup error: /root/PixiEditor/src/PixiEditor.Desktop/bin/Debug/net8.0/linux-arm64/libSkiaSharp.so: undefined symbol: uuid_parse`
The compilation environment is Debian 10 ARM64. I hope to get a version that can run in this environment. I am unsure whether it is due to the lower version of Debian 10 or the special architecture of ARM64. I am also unsure whether it can be compiled on Debian 10 x64. In any case, I hope someone can help me.

### Steps how to trigger the issue

```bash
1.
2.
3.
...
```

### How did you download the software?

PixiEditor Website

### Version Channel

Stable

### Version Number

master

### Operating System

Linux

### OS Version

debian10 arm64

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kero990:
I found that in `src/PixiEditor/Models/IO/Paths.cs`, `Assembly.GetEntryAssembly().Location `can return null or an empty string in my environment, causing path concatenation to fail. Replacing it with `AppContext.BaseDirectory` fixes the issue. It shouldn’t occur on x86_64, but `AppContext.BaseDirectory` is the officially recommended way to obtain the application base directory, so I submitted a PR.

Of course, since the official release does not include an arm64 Linux package, this change may not be merged, so I’m leaving a quick note here for anyone who wants to build the project on arm64 Linux.

PR #1313, opened 2025-12-30
Fix crash on dotnet publish by using AppContext.BaseDirectory

kero990:
Assembly.GetEntryAssembly().Location can return null or empty string in certain publish scenarios (single-file publish, some Linux deployments like aarch64), causing path combination to fail. Replace with AppContext.BaseDirectory which is the recommended way to get the application base directory in .NET Core/.NET and works reliably across all publish configurations.

Fixes [#1145](https://github.com/PixiEditor/PixiEditor/issues/1145)

### DIFF
--- a/src/PixiEditor/Models/IO/Paths.cs
+++ b/src/PixiEditor/Models/IO/Paths.cs
@@ -8,10 +8,10 @@ public static class Paths
     public static string DataResourceUri { get; } = $"avares://{typeof(Paths).Assembly.GetName().Name}/Data/";
 
     public static string DataFullPath { get; } =
-        Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), "Data");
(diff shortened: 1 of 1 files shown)


flabbet:
This seems good, although there are other places in the codebase that use Assembly.GetEntryAssembly(), as far as I remember FFMpegAnimationRenderer, but likely not limited to. Could you look into that?

flabbet:
@kero990 Could you confirm if other places are affected too? Otherwise I'll have to close the PR

kero990:
sorry—Claude helped locate this issue and the PR I’m not well-versed in C#, so I can’t vouch for the fix’s quality. Please feel free to handle the PR in any way you see fit; even if it only gives you a rough idea for a solution, I’d still consider that worthwhile.

flabbet:
I'll close the PR for now as I have more pressing things to implement and this requires more testing. Once we'll be adding ARM builds, I'll come back to this. Thank you for investigation!

PR #1366, opened 2026-02-09
Create cel to the right

flabbet:

Creating a new cel will now shift all other cels to the right, instead of finding an empty space and placing a cel there

## SELECT BELOW TO CONTINUE


CPKreu (changes requested):
3 issues I found

1. When creating a new frame for a layer that does not have a timeline yet, it starts on the second frame (Repo: Create document > Create new frame)
2. Some files appear to have random frames shifted to the right when reopening. See attached file, old Version the frames have no gap and this one creates a gap between the last 2 frames
3. You can stack multiple frames in the same time point (Repo: Create two frames > Put last frame at Position 1 > Move frame back to position 2)

And also had some code comments

CPKreu:
[Sheep moving frame.zip](https://github.com/user-attachments/files/25215007/Sheep.moving.frame.zip)
The missing attachment

flabbet:
Try now, it should be much better

PR #1388, opened 2026-02-27
Advisors and rasterize nested layer option

flabbet:

_Include the link to the related issues if relevant_ 

https://github.com/user-attachments/assets/a3312c08-0e53-4c7d-b96c-49cbb22af5db

## SELECT BELOW TO CONTINUE

### DIFF
--- a/src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/NestedDocumentNode.cs
+++ b/src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/NestedDocumentNode.cs
@@ -316,10 +316,13 @@ public class NestedDocumentNode : LayerNode, IInputDependentOutputs, ITransforma
 
     private void RenderNested(RenderContext ctx, Canvas workingSurface, Paint paint)
     {
-        if (NestedDocument.Value is null)
+        if (NestedDocument.Value is null || NestedDocument.Value.DocumentInstance is null || workingSurface is null || Instance is null)
             return;
 
         using var intermediate = Texture.ForProcessing(workingSurface.Surface, Instance.ProcessingColorSpace);
+        if (intermediate is null)
+            return;
+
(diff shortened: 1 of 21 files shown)


src/PixiEditor/ViewModels/SubViewModels/AdvicesViewModel.cs
```diff
@@ -0,0 +1,95 @@
...
+    const string OpenGraph = "OpenGraph";
+
+    public AdvicesViewModel(ViewModelMain owner, IAdvisor advisor) : base(owner)
+    {
+        Advisor = advisor;
+    }
+
+    public void RegisterAdvices()
+    {
+        if (!IPreferences.Current.GetLocalPreference<bool>($"Advice{NeedsNewLayerNestedAdviceKey}Dismissed"))
```
CPKreu: I think you could also use `PixiEditor.Extensions.CommonApi.UserPreferences.Settings.LocalSetting` here, so nobody accidentally uses different preference names for setting/getting

CPKreu (changes requested):
Interesting concept, I like that it's Mushy!

---

May I propose a instant close button on the Node Graph Mushy when you don't need the follow up

It felt like the Windows Setup Privacy screens, used it so many times but you had to click and read through it every time

---

And I'm not sure how I feel about typing out the text, I think showing it completely at once is also good? Just my opinion tough

src/PixiEditor/ViewModels/SubViewModels/AdvicesViewModel.cs
flabbet: I generally wanted to make this a generic function. Adding a hardcoded localsetting for each advice would be bloaty I think

flabbet:
> May I propose a instant close button on the Node Graph Mushy when you don't need the follow up

It's literally 2 sentences shown only once

Issue #1350, opened 2026-02-03
[BUG]: Shape colorspace shenanigans


bajsucks:
### Bug Description

Colorspace shenanigans in:
- unfragmented parts of raster rectangles (both pixel art and regular raster)
- raster circles (interestingly enough, pixel art circles display correct color)
- vector circles
- vector rectangles (only when snapped)

Color used in the demo is hex #341313FF, turns into #351616FF

demonstration:

https://github.com/user-attachments/assets/5f528fff-91e0-4773-a6a8-bea30372b178

### Steps how to trigger the issue

```bash
1. rectangle w/ hex #341313FF
2. 
3.
...
```

### How did you download the software?

PixiEditor Website

### Version Channel

Development

### Version Number

2.1.0.10

### Operating System

Windows

### OS Version

Windows 11 IoT Enterprise LTSC

### Additional Context

seems complex

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

flabbet:
you have no idea what rabbit hole you've just thrown me into

PR #1390, opened 2026-03-03
ChunkyImage is now using sRGB surfaces only

flabbet:

very scary change. From my tests, it seems to work.

https://github.com/PixiEditor/PixiEditor/issues/1350

## SELECT BELOW TO CONTINUE


flabbet:
@CPKreu If you have any .pixi files, just check if they look correct with this PR. Would be great if you had files from 2.0 and previous dev versions of 2.1

PR #1479, opened 2026-05-04
[FEATURE] EyeDropperTool: cross cursor and color preview overlay

MellKam:
## Problem

I found it a bit confusing that right now when you switch to color picker tool nothing changes on the screen, so for a minute I thought that something broke. The actual interaction (clicking the canvas to pick a color) wasn't discoverable.

https://github.com/user-attachments/assets/19e57241-db8c-48bf-88d0-dfab45de406c

## Changes

- Crosshair cursor — the cursor now switches to a crosshair when the color picker tool is active, which gives the user instant feedback that you can now select something.
- Color preview overlay — hovering over the canvas now shows a floating overlay with a color preview rectangle and the hex value of the color.

## Implementation notes

The original approach fetched a single pixel from the GPU buffer on each click. Extending this to fire on every mouse move turned out to be too slow and caused noticeable jitter. After trying a few throttle-based approaches I picked snapshotting the canvas buffer from GPU to CPU memory at the moment the color picker tool is activated, then doing simple index lookups on that cached copy during mouse move. So the result feels smooth and snappy.

One open question: should we invalidate that cached snapshot if the canvas is modified while the color picker is active? My assumption is that you can't edit the canvas in this mode, but I may be missing some edge cases. I'm open to any help on this

Because I don't normally write c#, I used claude for most of the code edits. Even tought I tried to review it but I can't guaranty that every bug is caught.

## SELECT BELOW TO CONTINUE


flabbet (changes requested):
Hey, thank you for this PR! The visual effect is really good, I like it. However there is one issue with the snapshot approach. If you have a color picker selected as a tool, and run an animation, the cache is stale and colors are invalid. 

There could be some other cases with custom graphs, like using `Time` (which can change the canvas without animation being enabled) - but I wouldn't worry about it if the color picker is in Single Layer mode. For Canvas mode, you can take advantage of [SceneRenderer](https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/Models/Rendering/SceneRenderer.cs) which takes care of rendering caching and holds previously rendered viewports. So no need to rerender document at all. Both SceneRenderer and Scene Textures are held in the [DocumentViewModel](https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/ViewModels/Document/DocumentViewModel.cs) class

Also, your solution does not take into consideration custom render outputs, I am attaching a file. But this again can be solved with `SceneTextures` in Document View Model.

(you can access the second output with wrench icon)

Image

[ColorTest.zip](https://github.com/user-attachments/files/27364741/ColorTest.zip)

MellKam:
@flabbet Hey. I rewrote it just like you said and now it seems to work with custom render outputs.

MellKam:
You might have noticed that in previous video when I switched from color picker -> brush -> color picker, the initial color shown was color of the previous cursor position(before tool switch). 

I found that this can be fixed by moving `overlay.PointerPosition = lastMousePositionOnCanvas;` to run unconditionally for every overlay, before the visibility check. Now all overlays are always in sync with the actual cursor position, so when the color picker overlay becomes visible it immediately has the correct current position.

https://github.com/PixiEditor/PixiEditor/pull/1479/changes#diff-8df576dbcd78084782e4c66b90f0ad2a862853b780f8dd5ed876344a832eeac6R517-R529

MellKam:
The thing I also noticed is that when you hold the left mouse button while using color picker tool with whole canvas scope it starts to lag pretty badly. Though with "single layer" it works smooth.

The problem was because every single mouse-move pixel triggered a full re-render: it allocated a new `Surface(documentSize)`, ran the entire node graph through RenderDocument(), then read one pixel from the result. 

Instead of re-rendering to pick a color, we can just read a pixel directly from the cached texture from `DocumentViewModel.SceneTextures`. So I did expose the active viewport's GUID via `IWindowHandler`, then replicated the `ViewportOverlays` fast path in `PickColor`.

https://github.com/user-attachments/assets/a9def678-1e6c-4b2c-a15f-d2464a52b205

flabbet (approved):
Looks good to me! Thank you.

I allowed myself to push a fix related to reference layer visibility, it was not introduced by you, but I included the fix in this PR for convieniance

PR #1485, opened 2026-05-06
Add aseprite import/export.

incomplete-tree:

title

Round trip video(don't recommend watching):

https://github.com/user-attachments/assets/68fb2210-e352-4ffa-894b-59b200d88ed5

## SELECT BELOW TO CONTINUE


src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+            }
+        }
+
+        // Build palette lookup
+        byte[][] palette = BuildPalette(paletteChunk, oldPaletteChunk0004, aseFile.ColorCount);
+        SetSwatches(builder, palette);
+
+        bool hasAnimation = aseFile.Frames.Count > 1;
+
+        Debug.WriteLine($"[AsepriteImport] Frames: {aseFile.Frames.Count}, Layers: {layers.Count}, HasAnimation: {hasAnimation}");
```
flabbet: Remove

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+        BuildLayeredDocument(builder, aseFile, imageLayers, palette, hasAnimation);
+    }
+
+    private void BuildSingleLayerDocument(DocumentViewModelBuilder builder, AsepriteFile aseFile, byte[][] palette)
+    {
+        Surface surface = RenderFlatFrame(aseFile, 0, palette);
+        builder.WithGraph(graph =>
+        {
+            graph.WithImageLayerNode(
+                new LocalizedString("IMAGE"),
```
flabbet, line 88: There's no layer name in ase file?

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+                surface,
+                ColorSpace.CreateSrgbLinear(),
```
flabbet: Aseprite blends colors with Srgb, not SrgbLinear

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+
+                var nodeBuilder = graph.AllNodes[^1];
+                nodeBuilder.WithInputValues(new Dictionary<string, object>()
+                {
+                    { StructureNode.OpacityPropertyName, aseFile.LayerOpacityValid ? layer.Opacity / 255f : 1f },
+                    { StructureNode.IsVisiblePropertyName, layer.IsVisible },
+                    { StructureNode.BlendModePropertyName, MapBlendMode(layer.BlendMode) }
+                });
+                nodeBuilder.WithPosition(currentPosition);
+                currentPosition.X += 500;
```
flabbet: Use 250 instead of 500

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+                var (layerIndex, layer) = imageLayers[i];
+
+                // Render first frame for this layer
+                Surface layerSurface = RenderLayerCel(aseFile, 0, layerIndex, palette);
+                string layerName = layer.Name ?? $"Layer {i}";
+
+                graph.WithImageLayerNode(
+                    layerName,
+                    layerSurface,
+                    ColorSpace.CreateSrgbLinear(),
```
flabbet: Same as above, use `CreateSrgb()`

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+            }
+
+            layerNode.KeyFrames = keyFrames.ToArray();
+            animationData.KeyFrameGroups.Add(layerGroup);
+        }
+
+        animationData.DefaultEndFrame = Math.Max(1, imageLayers.Count > 0 ? 
+            aseFile.Frames.Sum(f => Math.Max(1, (int)Math.Round((f.FrameDuration > 0 ? f.FrameDuration : 100) / (1000.0 / animationData.FrameRate)))) : 1);
+
+        Debug.WriteLine($"[AsepriteImport] AnimationData: FrameRate={animationData.FrameRate}, Groups={animationData.KeyFrameGroups.Count}, DefaultEndFrame={animationData.DefaultEndFrame}");
```
flabbet: Remove debug

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
```diff
@@ -0,0 +1,501 @@
...
+        foreach (var g in animationData.KeyFrameGroups)
+            Debug.WriteLine($"[AsepriteImport]   Group NodeId={g.NodeId}, Children={g.Children.Count}");
```
flabbet: same

flabbet (changes requested):
Thanks for this! It is already looking pretty solid and as far as I tested it works.

I found a few cases that does not work correctly though.

No matter what export option user chooses, it always exports the same, so layers + cels.
I think this is fine for the default Image export or `Save As`.

But it only exports cel animations, meaning that procedural animations do not export as frames (open fire pixel sample and try exporting to .ase). It makes sense as aseprite only has cel animations, but I feel like whenever user chooses "Animation" option in the export dialog, it expects procedural animation to be present too. 

In that case I'd force rasterize every frame of the animation and treat it as a cel.

And there's the third export option, Spritesheet. I think it would also be a natural choice to export it to .ase with a single spritesheet layer. I know you could export as .png and import to aseprite, but I'd do that for the sake of consistency. 

So to summarize
- Image export and "Save As" option -> Export layers + cel animation
- Animation export -> Assume there's an animation between frame 1 and specified end time and just rasterize every frame into an aseprite cel.
- Spritesheet export -> Single layer with spritesheet inside .ase file

Different export configurations are stored in `ExportConfig`.

If you think my proposition is not good, I'm open for a discussion

src/PixiEditor/Models/IO/CustomDocumentFormats/AsepriteDocumentBuilder.cs
incomplete-tree: No, because there are no layers in this file. We could use the filename to create a layer name, but using a default seems more logical.

flabbet: Makes sense

flabbet (commented):
So, in this file, we have 2 layers with animations. It gets flattened into one single animation. Is this behavior intended? As far as I know, aseprite supports animation per layer
[anim-ase.zip](https://github.com/user-attachments/files/27962721/anim-ase.zip)

incomplete-tree:
I thought about including layers in the export, this is probably what most people expect, but pixieditor's complex node graph makes this hard. What if someone connects layers via another node? Or what if someone connects layers to output indirectly ? 
The current system of taking the output and exporting it will always work regardless of the .pixi file, and i think this is the most "fair" way to export. 
If you don't agree, i could do something like ignoring any procedural art when there are multiple layers.

flabbet:
It's a valid point. Although I feel like it's easy enough to detect if graph only consists of Image Layer Nodes (and folders if aseprite supports them?) and if so, exporting per-layer would not be that hard and I believe it solves the issue.

- So if there's any other node in the graph, than Image Nodes, flatten like you did,
- If not, we can export all layers

DocumentViewModel contains methods of getting structure nodes in order, so it shouldn't be a problem

flabbet:
@incomplete-tree Hey, do you intend to change discussed behavior? No rush, I am just asking

incomplete-tree:
Yes, I just haven't gotten around to it yet.

PR #1489, opened 2026-05-06
Rainbow and Game of Life Brushes

Ghoti-tpt:
my first brushes yay

Adds a brush which cycles between colors over time (rainbow) and a brush that does conway's game of life to affected pixels.

https://github.com/user-attachments/assets/bcafae8e-a2d5-477e-b6df-c25037b66863

https://github.com/user-attachments/assets/55cc0e70-1502-460d-9d52-4a9c65c35ad4


Ghoti-tpt:
uhhh i guess the checkboxes i was supposed to fill out got weirded out because I made pr through very unconventional methods, apologies

flabbet (changes requested):
Thanks for the PR! I have a few requests though

## Rainbow

1. Please add all texts to translations, so SPEED, SATURATION and RATIO. Also change the brush name from Rainbow to RAINBOW_BRUSH and do the same.

Here's the file:
https://github.com/PixiEditor/PixiEditor/blob/master/src/PixiEditor/Data/Localization/Languages/en.json

2. Add tags to the brush (tags are located in the Brush Output Node)
3. Higher speed value = slower change? A bit unintuitive, I'd flip that
4. Preview should show a singular stamp instead of whole stroke since it's very small. Could you maybe make a circular gradient as a stamp? Remove the white background as well.
5. Unless you have a specific reason, please set spacing, it accepts values in 0-100 range, usually 1-15 is good enough. Even though there's no blending, It improves performance of the brush

## Game of Life

I will put this brush as a sample file instead of an actual brush, could you please change the layer so it has some starting pixels which brush can be used on? Additionally, include a vector text with an explanation how to use it

Ghoti-tpt:
Added the requested changes, also created a Find and Replace tool because it seemed pretty useful.
Added localization for that one too
Not sure what format to use for the Game of Life for making it into an example file so I left it as-is, though I did make the requested changes for that file too.

Ghoti-tpt:
as i have seen someone else already produce a better version of find and replace, i have removed it from this PR.

flabbet:
@Ghoti-tpt Who produced better version?

Ghoti-tpt:
@flabbet it was Laxan, my brush is solid color only which is objectively less useful, also goes outside the scope of this PR ngl
image

Laxan3000:
> @flabbet it was Laxan, my brush is solid color only which is objectively less useful, also goes outside the scope of this PR ngl
> image
> 

To be fair, a simple "Find and replace" brush could have been useful too, for the use cases of some people making pixel art and wanting to replace a color with another (maybe that brush could be improved to mimick the shade functionality of aseprite?)

flabbet:
Ah I see, one more thing, would you mind putting Game Of Life brush to `BetaExampleFiles` folder? https://github.com/PixiEditor/PixiEditor/tree/master/src/PixiEditor/Data/BetaExampleFiles

I feel like it would be better to be as an example file

Ghoti-tpt:
@flabbet Game of Life is now in example files

Ghoti-tpt:
@flabbet should I put my comment nodes + basic Find and Replace into this PR or a new one?

flabbet (approved):
lgtm. For new things, make a separate pr pls

PR #1525, opened 2026-05-23
Find and Replace and Swap Color tools, Comment and Stickynote Nodes

Ghoti-tpt:
Feel free to change any of the icons I used, couldn't think of anything better lol

Comment Node:
-Name and Text input fields.
-Resizable, recolorable and draggable highlighted area.

Stickynote Node:
It's just a very basic node with a single text field.

Find and Replace tool:
-Replaces pixels of primary color with secondary color.
-Adjustable Tolerance.

Swap Color tool:
-Changes the color of affected pixels to the primary color.
-MANY settings lol

## Comment and Stickynote Nodes

Usage of both

https://github.com/user-attachments/assets/e6c4a125-3a80-42b7-a407-7ec646c3bfe9

Changing color of Comment Node zone

https://github.com/user-attachments/assets/882fd860-da84-4548-a1d7-0d1d779456a6

## Find and Replace

https://github.com/user-attachments/assets/b1423f6e-717a-419a-906a-b6eb012f183b

## Swap Color

https://github.com/user-attachments/assets/5dc3e913-1e77-4eec-9da1-b8435b3fdae3

(thanks Laxan3000)

## SELECT BELOW TO CONTINUE


src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/StickynoteNode.cs
```diff
@@ -0,0 +1,19 @@
+using PixiEditor.ChangeableDocument.Rendering;
+namespace PixiEditor.ChangeableDocument.Changeables.Graph.Nodes;
+
+[NodeInfo("Stickynote")]
+public class StickynoteNode : Node
```
flabbet: Please rename it to `StickyNoteNode` (capitalize N)

src/PixiEditor/Data/Localization/Languages/en.json
```diff
@@ -1506,5 +1506,17 @@
...
+  "FIND_AND_REPLACE_TOOL": "Find and Replace",
+  "FIND_AND_REPLACE_TOOL_TOOLTIP": "Pixels of primary color will be replaced with secondary color. Tolerance adjusts color detection threshold.",
+  "FIND_AND_REPLACE_TOOL_ACTION_DISPLAY_DEFAULT": "Click on the canvas to replace colors.",
+  "COMMENT_NODE": "Comment",
+  "COMMENT_WINDOW_COLOR": "Box Color",
+  "SATURATION_TOGGLE": "Saturation Mode",
+  "BRIGHTNESS_TOGGLE": "Brightness Mode",
+  "SINGLE_COLOR": "Single Color Mode",
+  "SWAP_COLOR_TOOL": "Swap Color",
+  "SWAP_COLOR_TOOL_TOOLTIP": "Changes the colors of the affected area to match Primary Color. Use the checkboxes to preserve different color attributes. Single Color mode swaps only the color you click on. Hardness 100 should be used with low Tolerance.",
```
flabbet: This is not the purpose of tooltip. Use action display for this information. Tooltip should just contain short information about what the tool is.

src/PixiEditor/Data/Localization/Languages/en.json
```diff
@@ -1506,5 +1506,17 @@
   "STRANDS_BRUSH": "Strands",
   "WARNING_VERSION_INCOMPATIBLE": "This extension may not be compatible with your version of PixiEditor.",
   "EXTENSIONS_ONBOARDING_TITLE": "Featured Extensions",
-  "GAME_OF_LIFE_EXAMPLE": "Game of Life"
+  "GAME_OF_LIFE_EXAMPLE": "Game of Life",
+  "STICKYNOTE_NODE": "Sticky Note",
+  "FIND_AND_REPLACE_TOOL": "Find and Replace",
+  "FIND_AND_REPLACE_TOOL_TOOLTIP": "Pixels of primary color will be replaced with secondary color. Tolerance adjusts color detection threshold.",
```
flabbet: Tooltip should be `Finds and replaces pixels of primary color with secondary color` and
`Tolerance adjusts color detection threshold` should be in action display

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CommentNode.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public InputProperty<string> CommentName { get; }
+    public InputProperty<string> CommentText { get; }
+    public InputProperty<VecI> Size { get; }
+    public InputProperty<VecI> Offset { get; }
+    public InputProperty<Color> Color { get; }
+
+    public CommentNode()
+    {
+        CommentName = CreateInput(NamePropertyName, "NAME", "");
```
flabbet: What's the purpose of this property?

flabbet (changes requested):
I really like all of the new things in this PR, I have just a few nitpicks.

## Comment zone

I feel like the zone should not be separated from the node itself. Besides I think there's no benefit of having input properties connectable too. It would be much better if the node + zone would be a "one" UI element. I imagine it could be just a box you have now + a text above which is an editable comment.

@CPKreu Let me know what you think about it as well

## Sticky notes

It begs for a custom UI, like https://docs.unity3d.com/Packages/com.unity.shadergraph@9.0/manual/Sticky-Notes.html or at least a big text box (also it does not need a connection dot)

## Swap color tool

Add configurable shape

## Find and replace

The settings are a bit too confusing at the moment, idk if it's for you or laxian. I know that there's no way to make a list of options in the UI and having multiple checkboxes is a workaround, but maybe it is possible to make it more clear what things do. Or at least action display could describe it.

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/CommentNode.cs
Ghoti-tpt: To be honest, comment name is completely redundant so yeah I will remove it

CPKreu:
# Comment zone

I agree with flabbet about the one-UI thing. And not sure if this is what flabbet meant, but I'd think the node would be invisible and only the box has actual UI.
This should make it compatible with .pixi

An even better approach imo would be to somewhat separate the concept, where you would be able to put Nodes in a Frame
Basically the same concept as [Frames](https://docs.blender.org/manual/en/latest/interface/controls/nodes/types/layout/frame.html) from Blender.

I think for now with a invisible node it's fine, and putting Nodes in a Frame could be something we do in the future

# Sticky Notes

Pretty much what flabbet said, and would say it should be a invisible node too for .pixi compatibility and readability

---

In general I'd say these elements should be distinct from nodes in presence to make a node graph easier to read
Similar to how comments are usually easily distinguishable from code in programming languages

Ghoti-tpt:
New Comment and Stickynote
https://github.com/user-attachments/assets/88e65207-53b7-4a5a-87be-95ff2e104f42

flabbet:
It's much better with custom UIs. I think the look could be further improved, maybe @axinlab can suggest something

axinlab (changes requested):
### Comment Node:
1. I made a simple prototype of how the UI could be improved
- Color picker is visible only when editing the comment or the box is selected
- The text is centered, size the same as Node titles

https://github.com/user-attachments/assets/99fdc690-eb3b-435e-995a-cc6f0c711356

2. Text should wrap if too long (max width being the width of box below)
3. I also tested that atm if you minimize the box so that it disappears, there is no way to select the comment and delete it. With the design I proposed I think the min-height should be one line of text and width of the place holder

### Stickynote Node:
1. I'm missing a resizing option, just like the comment (with the min size too).
2. I don't think the title is needed?

I'd make it extra simple:

Image

In both proposals there is a problem with text color... Can we automatically change to white or black based on contrast?

I believe these changes improve readability of nodes, but let me know what you think.

Ghoti-tpt:
@axinlab @flabbet how does this look? (ignore the rendering issues near the end)

https://github.com/user-attachments/assets/7984b807-130e-4afa-9e98-9394da098a11

axinlab:
I like it! I'd test in on my machine before approving though. I don't see the changes here?

axinlab:
I'm also thinking about placement of the new nodes. 

Shouldn't the **Blackborad Variable Value** and **Document Info** be under **Inputs** category? @flabbet 

Then the new nodes could be a new category, smth like **Annotations** or **Notes**

image

Ghoti-tpt:
yeah they are there now, i just wanted to have those changes reviewed before i pushed them lol

flabbet:
it looks much better, 
1. How about adding rounded corners to comment?
2. Right now sticky note requires double clicking to edit the text, make it work like comment so it is focused on single click.
3. Nodes have shadows, maybe sticky note should too?
4. Min height of the comment and sticky note cuts the text, I think it should be minimum 1 line of text high
 
obraz

flabbet (approved):
lgtm! Thank you

flabbet:
@axinlab

axinlab (approved):

PR #1526, opened 2026-05-23
Changed path of autosave to be in local appdata

flabbet:

Tmp is not a good place for autosaves. It completely doesn't work on flatpak and might behave unexpectedly on some setups. Having them in localappdata solves these problems.

## SELECT BELOW TO CONTINUE

### DIFF
--- a/src/PixiEditor/Helpers/AutosaveUtility.cs
+++ b/src/PixiEditor/Helpers/AutosaveUtility.cs
@@ -4,6 +4,8 @@ namespace PixiEditor.Helpers;
 
 internal class AutosaveHelper
 {
+    static string legacyAutosavePath = Path.Join(Path.GetTempPath(), "PixiEditor", "Autosave");
+
     public static Guid? GetAutosaveGuid(string? path)
     {
         if (path is null)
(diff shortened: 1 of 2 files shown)


src/PixiEditor/Helpers/AutosaveUtility.cs
```diff
@@ -15,6 +17,13 @@ internal class AutosaveHelper
...
-        return Path.Join(Paths.PathToUnsavedFilesFolder, $"autosave-{guid}.pixi");
+
+        string path = Path.Join(Paths.PathToUnsavedFilesFolder, $"autosave-{guid}.pixi");
+        if (!File.Exists(path))
+        {
+            return Path.Join(legacyAutosavePath, $"autosave-{guid}.zip");
+        }
+
+        return path;
     }
```
CPKreu, line 29: How about doing a migration, instead of reusing the same path, so we can reduce this code again in the future?

flabbet: I think we can already safely reduce this code in the future as no new autosaves will be in legacy autosave path

CPKreu: Ok right, yeah make sense

CPKreu (approved):

PR #1522, opened 2026-05-22
UI polish and workflow improvements

Gedudo:
Hi, 
This is a PR dedicated to adding some UI polish and minor workflow improvements that I started working on before realizing the contribution guide specifically asks that new features be discussed first or be based on existing issues. Apologies for not bringing this PR up earlier and discussing it first, I am however hoping that perhaps discussion could be had on the work I was able to do in this PR. I'm completely open to this PR being taken in a different direction or being modified, its purely just for review and the maintainers can use it however they feel fit and whichever way they would find it most useful.

**Disclaimer:** AI assistance was used in making this PR and I am not a proper developer 

Overview screenshot of the existing 2.1.1.4 UI
(Overview screenshot of the existing 2.1.1.4 UI)

Overview screenshot of the entire modified UI
(Overview screenshot of the entire modified UI)

 ## Clearly describe changes, as detailed as possible

### UI Polish

- Converted all fonts that aren't monospace to Inter Tight for a sharper look and a bit of familiarity to other FOSS projects
- Made all text in the footer the same size and moved the coordinates and version to the bottom right corner
- Removed pen icon on news posts and shifted date to left side
- Centered the play button in the title bar to the width of the window
- **Added hover fill when hovering over buttons**
  - Selected ring is now grey instead of white
- Reworked Layer Outliner
  - Added icons for layer types and moved/scaled folder icon to fit styling
  - Layers that are hidden individually or hierarchically will be greyed out
  - Resized "Add Reference Layer" button and internal elements
  - Reduced the size of the icons for buttons like add layer, add folder and etc

### Workflow Improvements

- Moved toolset/ mode selector into toolbar dropdown
- Reworked tool bar Font Dropdown. 
  - Added more styling and 
  - Added live font previewing for selected text in the viewport
- Added a width/height swap button to the "Create new image" page
- Added background color toggle and color selector to the "Create new image" page

 ## If possible, show examples of usage

(Showcase of the modified news feed, new swap width/height button and new set background color toggle and color picker)
PixiEditor Desktop_ec1oMswnHO

(Showcase of working greyed-out hiding functionality)
PixiEditor Desktop_qHmCaHGDMS

(Showcase of the new fill when hovering over buttons, the new dropdown mode/toolset switcher and the font dropdown + live preview)
PixiEditor Desktop_4191WfLZJx

## SELECT BELOW TO CONTINUE
- [ ] I wrote tests for my changes (if possible)
- [ ] I've included XML docs inside the code in relevant places
[...]

PR #1528, opened 2026-05-23
Improved Layer UI

Gedudo:
- Added icons for layer types and moved/scaled folder icon to fit styling
- Layers that are hidden individually or hierarchically will be greyed out

From original PR: https://github.com/PixiEditor/PixiEditor/pull/1522

Showcase

## SELECT BELOW TO CONTINUE


axinlab:
I love that improvement! I think it would be worth adding tooltips to icons on hover:
- Vector Layer
- Raster Layer
- Layer Group

The same style as for the tools in toolset. What do you think?

Gedudo:
Thank you! Yeah i think that would be a great addition, Ill definitely look into into adding it!

Gedudo:
> I love that improvement! I think it would be worth adding tooltips to icons on hover:
> 
>     * Vector Layer
> 
>     * Raster Layer
> 
>     * Layer Group
> 
> 
> The same style as for the tools in toolset. What do you think?

Ok I have added the tooltips in!

Tooltips Example

axinlab (changes requested):
After looking a bit more into layers I noticed there are more types. Here are (I hope all) listed:

Image

I also made new icons, I think they are more recognizable:
**Raster Layer** 
Image
**Vector Path**
Image

Gedudo:
The icons are a huge improvement!! 

Ill definitely implement them. I questioning the implementation of some of these different layer types. I definitely think the Text layer type should be implemented and Nested Document if its stuff like imported media and whatnot. However Line, Ellipse and Rectangle I think might be a poor choice to add as they seem to be both raster or vector dependent from messing around with them in the program as well as being able to use other raster or vector tools in the same layer such as the pen tool or vector paths tool. In general im not sure these tools should be making a new layer unless you are working in a different toolset from what the layer supports such as working with a vector toolset on a raster layer.

Also a separate thing from further messing around with the layer outliner, im thinking there should be a warning if a vector layer is merged into a raster layer which warns the user that the vector layer will be rasterized with an additional checkmark box which doesn't show the warning again. This however im not completely sure if its in scope of this PR and I think I might need some support if this were to be implemented as I don't want to break anything by accident.

axinlab:
> However Line, Ellipse and Rectangle I think might be a poor choice to add as they seem to be both raster or vector dependent from messing around with them in the program as well as being able to use other raster or vector tools in the same layer such as the pen tool or vector paths tool.

For the Painting / Pixel Art toolsets, each new shape, stroke, text etc is added to the selected layer and there is no way to edit them as objects later. The Vector toolset works on objects, meaning that each shape, text or path is a separate layer. So the icons of vector shapes would be useful for example if the user changes a shape name. The icon would indicate it is an object that can be edited.

There is no way to mix raster and vector in the same layer, you'd have to rasterize a vector object to merge two layers together (I think a warning when rasterizing a vector shape could be a nice addition, but for a different PR).

Gedudo:
Apologies for the lack of response to this. I will look into implementing the new icons and additional layer types when I have the chance! I also appreciate the clarification.

flabbet:
Closing for inactivity. Let me know to reopen

PR #1533, opened 2026-05-24
New dropdown picker style

Gedudo:
## Clearly describe changes, as detailed as possible

This PR simply adds a style for a dropdown picker that these PRs rely on:
- Toolset selector placement change: https://github.com/PixiEditor/PixiEditor/pull/1534
- Font picker change: https://github.com/PixiEditor/PixiEditor/pull/1535

From original PR: https://github.com/PixiEditor/PixiEditor/pull/1522

## SELECT BELOW TO CONTINUE
- [ ] I wrote tests for my changes (if possible)
- [ ] I've included XML docs inside the code in relevant places
- [x] I agree to [PixiEditor's Code of Conduct](https://github.com/PixiEditor/PixiEditor/blob/master/CODE_OF_CONDUCT.md)

PR #1534, opened 2026-05-24
Toolset selector placement change

Gedudo:

- Moved toolset selector to toolbar

Relies on to function: https://github.com/PixiEditor/PixiEditor/pull/1533

From original PR: https://github.com/PixiEditor/PixiEditor/pull/1522

### Current PR:

Actual Example

### Alternate style concept:

Concept Example 1

Concept Example 2

## SELECT BELOW TO CONTINUE


axinlab:
After looking into this with @flabbet, we think this part of the interface deserves a proper redesign. I'll test a few different directions with users first to make sure we pick the best one.

Closing this for now, but thanks for the work here!

PR #1564, opened 2026-06-05
Change the CPU representation of Half4 from Color to Vec4D

### LINKED ISSUE #1531: [BUG]: Separate Color HSV/HSL broken
### Bug Description

Hello!

Separating colors is broken for everything but RGB, showing completely different values on the output side than the input side.

The Combine Color node seems fine, at least based on my testing with raw doubles as inputs.

Image

### Steps how to trigger the issue

```bash
Attach a Color Node to a Separate Color Node.
Select HSV/HSL
Route ouputs into equivalent Combine.
```

### How did you download the software?

GitHub

### Version Channel

Stable

### Version Number

2.1.1.4

### Operating System

Linux

### OS Version

Zorin OS 18 (Ubuntu 24.04.3), GNOME X11

### Additional Context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Equbuxu:

Fixes #1531
Drawie PR for this change: https://github.com/PixiEditor/Drawie/pull/9

## SELECT BELOW TO CONTINUE


Equbuxu:
This should be ready now but 

- There is a slight issue, when you hover over the Color connector it doesn't show it as color, instead it shows the underlying vector. I can't think of a good way to fix this, since it simply calls `ToString()` on `Vec4D`

image

instead of

image

Maybe it's fine to leave as is?

- `Vec4D`'s now have a hardcoded color picker as their `PropertyView` which might become a problem if we decide to use `Vec4D`'s for stuff other than colors. 
- More of a note, instead of making a new view for `Vec4DPropertyViewModel` I mapped it to `ColorPropertyView`, so `ColorPropertyView` now has two different view models
- Do I need to worry about `/PixiEditor.Extensions.Sdk/build/EncryptionLib.dll` and other ones being in the commit? rider did something to them

flabbet:
> There is a slight issue, when you hover over the Color connector it doesn't show it as color, instead it shows the underlying vector. I can't think of a good way to fix this, since it simply calls ToString() on Vec4D

I guess its' fine, although there should be a way to do detect if it's Vec4D used in color context

> Vec4D's now have a hardcoded color picker as their PropertyView which might become a problem if we decide to use Vec4D's for stuff other than colors.

It's fine, we can make it configurable on view model level

> Do I need to worry about /PixiEditor.Extensions.Sdk/build/EncryptionLib.dll and other ones being in the commit? rider did something to them

no

flabbet (changes requested):
It's not possible to connect Color to Shader's Color input anymore. Check `Basic.pixi` as a example file (It's in Data/Brushes)

Equbuxu:
I've added a converter which seems to have fixed what you mentioned but there are two new issues
- It turns out this PR accidentally fixed #1602, and the basic brush has a 255 entered into a combine color node, but it should be 1 if we are gonna make colors go from 0 to 1 in all cases. So I guess the basic brush needs to be updated, and this will break other documents too
- There is an issue with serialization/deserialization, the manually selected colors in color node in files created earlier no longer load in this PR, i guess because the type is different. Weird that it didn't throw

flabbet:
@Equbuxu I think we can just perform file conversion to fix it? Deserialization issue sounds fairly easily fixable too

flabbet:
+ render tests fail

flabbet (approved):

PR #1566, opened 2026-06-05
Improve localization debug window and fix various issues in it

Equbuxu:

- [x] Made it so the localization window itself is not localized (strings are hardcoded now). Reasons: the window isn't user facing and translators should know english anyway, the strings from that window were confusing and weren't correctly translated on POEditor, the window itself changing when the languages changes is a big jarring, and it's just simpler if we don't translate it. I've also deleted all the keys that were only used there.
- [x] Rephrased all the labels like "Source newer", "Update source" to be more understandable
- [x] Fixed the "Download language data and write it into .json" button (Previous known as "Update source")
- [x] Fixed `LastUpdated` field not being parsed from the built-in json due to being private (and therefore the language statuses being wrong)
- [x] Tweaked the margins a bit and added a border to the current language
- [x] Added a "quickly switch to previous language" button, I've found myself doing this manually a lot while translating strings

Old:
localization-debug-old

New:
localization-debug-window-new

## SELECT BELOW TO CONTINUE


Equbuxu:
Added a thing that hides the key and shows ••• instead

flabbet (approved):
very cool!

PR #1593, opened 2026-06-18
Fix Swap Color Brush

Laxan3000:
Changes:
- Made the brush easier to use by removing the Hardness, which made the brush more confusing to use;
- Fixed so the brush can draw better on transparent pixels.

## SELECT BELOW TO CONTINUE


flabbet:
Is this a correct behavior? I mean the color blob at the first stamp

https://github.com/user-attachments/assets/0d73d6b8-4918-46bd-975f-85582cf874ae

Here's a png

flower

Laxan3000:
I honestly wouldn't be able to tell if this is a brush issue or a pixieditor issue, because there's technically no difference in how the first stamp and later stamps are treated. Does this happen when "lock on target color" is off too?
That mode locks on a single color when dragging, instead of changing the target color while dragging, so for JPGs (or compressed PNGs with artifacts) it can kinda do that ig if the tolerance is high

flabbet:
As far as I remember it happens regardless of the brush settings, it wasn't there before so I'd say it's a brush issue. I can't merge it this way

-------- Original Message --------

PR #1607, opened 2026-07-02
Fixing arrays of arrays

Ghoti-tpt:

Arrays of Arrays didn't work before (just became null)
Also improved Value to String conversion so you can see the values of said arrays in their sockets

You can now make nested arrays.
image

Improved socket visualization (a lot of things that used to say null now show their values)
image

## SELECT BELOW TO CONTINUE
- [X] I wrote tests for my changes (if possible)


Ghoti-tpt:
oh yeah also now when you get an element (normal array) from a nested array it doesnt produce an extra Array node when you connect to a second Array Element node, like below

image

flabbet:
Can you make the pink socket somehow represent multi dimensional array? 

obraz

My suggestion is to draw it like normal array but maybe split in two? 

PixiEditor_73gMhFmub3

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/Arrays/ArrayConverterNode.cs
```diff
@@ -123,45 +117,6 @@ protected override void OnExecute(RenderContext context)
...
-
-            if (commonBaseType == null || commonBaseType == typeof(ValueType))
-            {
-                return typeof(object);
-            }
-        }
-
-        return commonBaseType ?? typeof(object);
-    }
-
```
flabbet, line 164: Couldn't find issues with this yet but I have a feeling replacing it with `Output.InternalProperty.ValueType.GetElementType() ?? typeof(object);` feels too easy. Could you explain a reason behind it and why it caused issues?

Ghoti-tpt: well when trying to default to the common base with 2 inputs for an array Array1 that is then attached to an array Array2:

(eg. 2 math node outputs 1.0 and 0.0) 
commonBaseType is a Func, typeof(Func<FuncContext, Float1>)
i = 1
currentType is a Func, typeof(Func<FuncContext, Float1>)
inputs[1].Value is del
val is a Float1
since it is a ShaderExpressionVariable it does GetConstant() -> 0.0, val is now a double
currentType is a double
so it compares a double and a Func, the only common base there is object

Array1.Output.Value is an object[] {1.0, 0.0}
Array1.Output.InternalProperty.ValueType is still Func<FuncContext, Float1>[]
Array2.First.InternalProperty.ValueType = Func<FuncContext, Float1>[], with Object[] value gets an exception and is caught in TryConvert -> false, so the value is null
Array2.First.Value = null, which makes Array2.Output.Value = [null]

the replacement just uses the declared element type, since it is already given by syncgroup which ensures the inputs and the output's element type are all the same type already

Ghoti-tpt:
multidimensional arrays now look like this
image
Screenshot 2026-07-03 151651

src/PixiEditor.ChangeableDocument/Changeables/Graph/Nodes/Arrays/ArrayConverterNode.cs
flabbet: Yeah, that makes sense. I'll do some more testing and merge if I find nothing

PR #1638, opened 2026-07-17
Iterative rendering

flabbet:

Significant optimization for structure-only graphs

## SELECT BELOW TO CONTINUE
- [ ] I agree to [PixiEditor's Code of Conduct](https://github.com/PixiEditor/PixiEditor/blob/master/CODE_OF_CONDUCT.md)


Equbuxu:
I've found #1640 and #1641 while testing this, but I can also reproduce them outside of this PR, so I guess they aren't related.
Apart from that it seems to be working well. I haven't spent enough time looking at the code to really properly understand the changes, and everything seems to be the same as before. I guess it would be nice to have some benchmarks to check if it actually improves performance (it sounds complicated to benchmark though), unless maybe it's obvious to you that it should or if you can clearly feel the difference. Some copilots suggestions look like they might be valid

Equbuxu (commented):
.

PR #1655, opened 2026-07-22
Add Background Eraser Brush

Laxan3000:
Add Background Eraser Brush
https://github.com/user-attachments/assets/e7b84d86-50cc-4d4b-beb0-2954899b8a60

Also fix a minor issue with swap color brush

# TODO
make an icon for it in the icomoon file with the name `bgeraser` (as the name `icon-bgeraser` in ToolSetsConfig.json)

## SELECT BELOW TO CONTINUE


flabbet (changes requested):
I feel like "Lock on Target Color" should be enabled by default. Also is there any use in having it disabled anyway?

Laxan3000:
> I feel like "Lock on Target Color" should be enabled by default. Also is there any use in having it disabled anyway?

Having it disabled makes it act more like the Background Eraser from Photoshop/Photopea (it erases the similar colors as you drag around your subject), having it enabled only filters the first color you clicked on as you drag (useful for backgrounds of the same color, as in the example I've shown)

Laxan3000:
Also a thing to note: both this brush and the swap color brush work best with pixelated shapes, as rounded ones makes for transparent pixels the shader has nothing to do about (I believe that's a pixieditor thing?)

flabbet (approved):

PR #1663, opened 2026-07-23
Delete MouseUpdateController

Equbuxu:

because it seemed to only slow things down

## SELECT BELOW TO CONTINUE


flabbet (approved):
looks good to me, I don't feel much difference but it definately seems good to get rid of this code

PR #1665, opened 2026-07-24
Change PR template to stop github from parsing checkboxes as tasks

Equbuxu:
I tested this in the description of #1663, apparently adding a second dash makes github think it's no longer a task. It also makes it show bullet points but I think that's fine

image

image

## SELECT BELOW TO CONTINUE


flabbet:
lgtm

PR #1666, opened 2026-07-24
Action accumulator work scheduling

flabbet:

Currently, when Brush Engine has to process tons of new points, it does it immediately, which can stall the render thread and cause rendering spikes. By adding action accumulator scheduling, we can set processing budget per frame and spread work across multiple frames. While it does not improve performance of individual work item by itself, it improves the feeling and smoothness of the app by removing spikes.

## SELECT BELOW TO CONTINUE


flabbet:
@Equbuxu would you mind testing? I made some CPU optimizations in the Brush Engine. I don't expect magic, but maybe something.

PR #1667, opened 2026-07-24
Optimize Blending Paint brush

Ghoti-tpt:

Great optimizations

-removed 16x eval calls which are wasted
-does flat color calculation since it doesnt vary based on sample uv
-spacing default set to 6% so runs 3x less often

doesnt lag my drawing tablet at all when making very long strokes

## SELECT BELOW TO CONTINUE


flabbet (approved):
lgtm, thanks!

PR #1670, opened 2026-07-26
Update shader node default code with version 300

Ghoti-tpt:

idk why this wasn't added yet lol, added `#version 300` to the top of the default shader node code

## SELECT BELOW TO CONTINUE


Equbuxu:
from what I was able to figure out myself, despite this being undocumented the #version 300 thing is implemented properly in SKSL, there is an internal version type in skia source code as well as some associated versioning logic. The bugtracker at https://issues.skia.org/ also has some issues with people talking about versioning. So it seems that the feature is solid enough that we can actually expect all the stuff enabled by `#version 300` to continue being supported?

The only thing that comes to mind, could it be that there are machines where PixiEditor runs but scripts with `#version 300` won't compile or run because of very old hardware/graphics drivers?

Equbuxu (approved):
works on my machine, I think this is fine to merge, if it does cause issues for someone (feels unlikely) we should find out soon enough

PR #1688, opened 2026-08-16
Animation editor fixes

Ghoti-tpt:

Fixed Animation Editor Issues:
-Pixel Perfect brushes would bug out in animation editor
-Frame Previews in the Timeline would be blank when duplicate button is pressed or frame length is changed until first edit to the texture
-Animation Editor was surprisingly laggy when manipulating frames

First 2 fixes:
https://github.com/user-attachments/assets/1320b68f-6c15-464b-ac9b-b958a7680f4d

## SELECT BELOW TO CONTINUE
- - [X] I wrote tests for my changes (if possible)


src/PixiEditor/Models/DocumentModels/UpdateableChangeExecutors/PenToolExecutor.cs
```diff
@@ -27,12 +27,11 @@ internal class PenToolExecutor : BrushBasedExecutor<IPenToolHandler>
 
     public override ExecutionState Start()
     {
+        pixelPerfect = GetHandler<IPenToolHandler>().PixelPerfectEnabled;
+
```
flabbet, line 31: What was exactly happening with brushes and how does this change affects it?

flabbet (approved):
Overall looks good to me. Let me know about this pixel perfect thing and we can merge

src/PixiEditor/Models/DocumentModels/UpdateableChangeExecutors/PenToolExecutor.cs
flabbet: nvm, got it

PR #1696, opened 2026-08-26
fix(ui): added a scrollbar to Blackboard and prevent it from clipping

ainthony:
When creating several blackboard variables that go beyond the monitor dimensions, it was previously impossible to scroll down and access them.
This change resolves the issue by:

Wrapping the ItemsControl inside BlackboardView.axaml in a ScrollViewer with VerticalScrollBarVisibility="Auto" and a bottom margin to prevent clipping against the status bar.

Adding a MaxHeight="{Binding $parent[UserControl].Bounds.Height}" constraint to the parent Border inside NodeGraphDockView.axaml to prevent infinite height expansion.

https://github.com/user-attachments/assets/f8a1023d-7f8e-4528-89a8-9da43b95b92a

The first blackboard uses the unedited blackboard that shows the variables going beyond the monitor dimensions. The next one shows the scroll wheel appearing when the amount of variables go beyond the monitor dimensions. 

## SELECT BELOW TO CONTINUE
- - [ x] I agree to [PixiEditor's Code of Conduct](https://github.com/PixiEditor/PixiEditor/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/src/PixiEditor/Views/Blackboard/BlackboardView.axaml
+++ b/src/PixiEditor/Views/Blackboard/BlackboardView.axaml
@@ -11,9 +11,9 @@
              x:Class="PixiEditor.Views.Blackboard.BlackboardView">
     <Grid>
         <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
             <RowDefinition Height="*"/>
-            <RowDefinition Height="Auto"/>
-            <RowDefinition Height="Auto"/>
         </Grid.RowDefinitions>
         <Grid.ColumnDefinitions>
             <ColumnDefinition Width="*"/>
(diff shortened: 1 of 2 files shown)


flabbet (changes requested):
Hey, thanks for the PR. It seems to work as expected, one request though. Previously, below the variable type, there was no empty space. In your PR there is a block of emptiness 

Image

ainthony:
I managed to fix the block of emptiness. 

https://github.com/user-attachments/assets/0b8a778d-d0fe-4ba8-80d9-b8bedbb3071e

flabbet (changes requested):
Block of emptiness is gone, but there is one more issue introduced by this PR. Tbh I have no idea why, it slightly looks like avalonia bug. You may be able to fix it with virtualized stack panel instead of a normal stack panel, but not sure.

I'll also merged the master branch into the PR, it upgrades avalonia to version 12 and .NET 10, so you might need to download .NET 10 sdk

https://github.com/user-attachments/assets/94751e02-d8fb-4507-afae-3f288f19dd83

ainthony:
ok I think i finally got it without any bugs this time 
https://github.com/user-attachments/assets/ededf55e-7eac-4e11-9cb4-f046c7bc8f25

flabbet (approved):
Seems to work now, thanks!

Issue #1682, opened 2026-08-12
[BUG]: The code in SeparateColorNode makes no sense, the node is broken when Normalized Color Values is unchecked


Equbuxu:
### Bug Description

i literally wrote it not too long ago and I don't know what I was thinking, for example `GetHsva` first tries to multiply the input channels by 360, 100, 100, and 255 in `AdjustForHsvRange` (so treats the input like it is HSVA) but after multiplication the result is fed into `RgbaToHsva`, so now we are treating the input as if it is RGB. I think `AdjustForHsvRange` and `AdjustForHslRange` aren't needed at all and should be replaced by `AdjustForRgbaRange` like it worked before.

Besides, logically `Normalized Color Values` should refer to the format of the outputs of the node, so we shouldn't even be looking at that checkbox before converting to target color space? At least I think that's what the checkbox is supposed to mean? Not actually sure

Here is what it manifests as:

<img width="612" height="250" alt="Image" src="https://github.com/user-attachments/assets/9cde0591-83be-4ec9-9fbb-5e43dbac04f6" />

### Steps how to trigger the issue

```bash
1.build node from screenshot
2.hover over outputs
3.see big numbers
...
```

### How did you download the software?

PixiEditor Website

### Version Channel

Stable

### Version Number

2.1.2.1

### Operating System

Windows

### OS Version

Win 11

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1709, opened 2026-08-31
Fixed Separate and color node with HSV and HSL

flabbet:

#1682

PR is not done yet as there are some cases that don't work in current release (yet the issue described above does not exist there)

## SELECT BELOW TO CONTINUE


flabbet:
@Equbuxu I am losing my mind, can you take a look at this?

I have no idea what is going on with case attached. CPU HSL -> CPU HSL works as expected. But in GPU context instead of red, we're getting cyan. The best part is, I did not change code related to GPU context at all, it must've break somewhere along the way, not with this PR. 

- Hardcoding all values - works
obraz
- Passing hue seems to be broken? Hardcoded value of 0 works
obraz
- All together - broken
obraz

Shader makes total sense
```sksl
uniform float2 iResolution;
layout(color) uniform half4 color_3;

... (truncated HueToRgb and HslToRgb)

half4 main(float2 coords)
{
   coords = coords / iResolution;
   float2 vec2_2 = float2(coords.x, coords.y);
   vec2_2 = coords;
   half4 color_4 = RgbToHsl(color_3);
   half4 color_5 = HslToRgb(color_4);
   half4 premultiplied = half4(color_5.r * color_5.a, color_5.g * color_5.a, color_5.b * color_5.a, color_5.a);
   return premultiplied;
}
```

Passing color uniform to shader also seems to work correctly, as plugging Color node to EndMofiyImage Color input gives correct result too

obraz

So it's not an issue of uniform input data, it's not an issue of combine color hsl algortihm, I don't have a proof that it's not an issue of RgbToHsl algorithm, but it was not changed since the working version and well... it does work there.

[opac.zip](https://github.com/user-attachments/files/31646396/opac.zip)

flabbet:
Hue value that perfectly matches this cyan is exactly `0.5`, which is neither H value of red in the sample (0) and not R value of this red (0.68). However I am pretty sure the shader itself is correct and the value that is passed to the shader is wrong, but debugging directly the uniform data that is to be passed to shader also gives correct value (0.68, 0.38, 0.38, 1)

PR #1719, opened 2026-09-14
Refactor node traversal methods

CPKreu:
Refactored node traversal methods for both backend and frontend in a single reusable `PixiEditor.GraphNavigation` library

- Changes to traversal are now directly reflected in frontend and backend
- There is now a single definition for backwards and forwards that uses a `TraverseContext` instead of multiple value tuples definitions

  ```cs
  public void TraverseBackwards(Func<TraverseContext<TNode, TInput, TOutput>, Traverse> func)
  
  public void TraverseForwards(Func<TraverseContext<TNode, TInput, TOutput>, Traverse> func)
  ```
- TraverseContext is more expressive than value tuples and provide documentation
- Both frontend and backend now use the same Traverse enum (only implemented in frontend previously) instead of a boolean indicating further or exit

# Example refactoring

```cs
        Content.Connection.Node.TraverseBackwards((n, _) =>
        {
            if (n is not StructureNode { IsVisible.Value: true } structureNode ||
                n is IReadOnlyFolderNode)
                return true;

            var childBounds = structureNode.GetTightBounds(frameTime);
            if (childBounds != null)
                bounds = bounds?.Union(childBounds.Value) ?? childBounds;

            return true;
        });
```

to

```cs
        Content.Connection.Node.Navigate().TraverseBackwards(ctx =>
        {
            if (ctx.Current is not StructureNode { IsVisible.Value: true } structureNode ||
                ctx.Current is IReadOnlyFolderNode)
                return Traverse.Further;

            var childBounds = structureNode.GetTightBounds(frameTime);
            if (childBounds != null)
                bounds = bounds?.Union(childBounds.Value) ?? childBounds;

            return Traverse.Further;
        });
```

---

Todo:

- [x] Write navigation library and implement in frontend and backend
- [x] Write wrapper extensions in frontend/backend to make calling simpler
- [ ] Write tests
- [ ] Refactor codebase to remove old methods


Equbuxu:
Could you please make better comments for what the entries in the `Traverse` enum mean, the current ones are confusing to me (they might or might not be correct, but "go further in this direction, child connections will not be enqueued" sounds contradictory?)

Also, I remember when i first came across the Traverse methods I thought at first that it's just DFS and that returning false would skip the current search branch, I also wasn't sure if the starting node is included in the traversal, these two things would also be good to document

CPKreu:
> Could you please make better comments for what the entries in the `Traverse` enum mean, the current ones are confusing to me (they might or might not be correct, but "go further in this direction, child connections will not be enqueued" sounds contradictory?)
> 
> Also, I remember when i first came across the Traverse methods I thought at first that it's just DFS and that returning false would skip the current search branch, I also wasn't sure if the starting node is included in the traversal, these two things would also be good to document

@Equbuxu 

1. Oh yeah, that comment is wrong. Will fix
2. The Enum already fixes (with proper comment) the boolean confusion, right?
    For the starting node, I'll add a boolean like this `Backwards(bool includeOrigin = true)` to make it really explicit without needing to look at comments

